PR #8, opened 2021-04-02
docs/variants/workstation/: installation manual initial version

pietrushnic:
Signed-off-by: Piotr Król <piotr.krol@3mdeb.com>


docs/variants/workstation/installation-manual.md
```diff
@@ -0,0 +1,218 @@
...
+
+    ![](../../images/disk_dock_open4.jpg)
+
+7. When the dock is removed the service mode jumper should be visible.
+
+    ![](../../images/overview_service_jumper.jpg)
+
+8. Place the jumper in the place marked by red rectangle.
+
+    ![](../../images/service_jumper.jpg)
```
mikebdp2, line 51: Rename this file to service_jumper_header.jpg

docs/variants/workstation/installation-manual.md
```diff
@@ -0,0 +1,218 @@
...
+
+9. It should look like this.
+
+    ![](../../images/image0.jpeg)
```
mikebdp2, line 55: After that, image0.jpeg could be renamed as service_jumper.jpg.

docs/variants/workstation/installation-manual.md
```diff
@@ -0,0 +1,218 @@
...
+active. Press F1 to proceed  and boot to your Linux system.
+
+    ![](../../images/service_mode_warn.jpg)
+
+## Installation
+
+You will need root privileges from now on to proceed. So switch to root user or
+use __*sudo*__ in each command.
+
+1. Install flashrom v1.1 with your distribution's package manager, if you don't
```
mikebdp2, line 67: v1.1 - why not v1.2 ? Also: if the success of installation could depend on a flashrom's version number, please provide the minimum flashrom version requirement. And point to the instruction like https://www.flashrom.org/Downloads in case the customer's flashrom version number provided by his Linux distro is insufficient: i.e. some ancient Debian could have a flashrom 0.x at repositories.

docs/variants/workstation/installation-manual.md
```diff
@@ -0,0 +1,218 @@
...
+   that Protected Range (PR) restrictions still apply.
+   ```
+
+   Newer version of flashrom may not display the warning about unsupported
+   chipset as it already may be marked as tested. Our team has verified that the
+   flashrom updates firmware reliably on this chipset.
+
+## Verification
+
+1. If everything went well (flashrom has verified the flash content),
```
mikebdp2, line 133: 1. ... :
2. Shut down **the** machine, move **a** jumper to the original place

docs/variants/workstation/installation-manual.md
```diff
@@ -0,0 +1,218 @@
...
+2. Shut down machine, move jumper to the original place
+3. Power on the machine.
+4. After rebooting you should see the Dasharo Workstation logo when booting.
+   When logo appears you may press ESC to select the boot device if you want to
```
mikebdp2, line 137: When **a** _Dasharo_ logo appears**,** you...
After finalizing the text, we should look at the articles and check if the punctuation could be improved to make a text more readable.

docs/variants/workstation/installation-manual.md
```diff
@@ -0,0 +1,218 @@
...
+
+``` console
+ERROR: Could not get I/O privileges (Operation not permitted).
+You need to be root.
+Error: Programmer initialization failed.
+```
+
+Use `sudo` with the flashrom command. If the problem persists, probably the
+kernel restricts access to IOMEM. In order to workaround it, append
+`iomem=relaxed` to kernel commandline. You may do it in two ways:
```
mikebdp2, line 174: There's a third, simpler way:
```
    If your computer uses BIOS for booting, then hold down the Shift key while GRUB is loading to get the boot menu.
    If your computer uses UEFI for booting, press Esc several times while GRUB is loading to get the boot menu.
```
and, after getting a GRUB menu, press "e" on a boot entry to append `iomem=relaxed`to kernel commandline and press Ctrl+x or F10 to boot. Although this setting is temporary and will last only during the next boot, this way is faster and a customer doesn't need to re-generate anything.

docs/variants/workstation/installation-manual.md
```diff
@@ -0,0 +1,218 @@
...
+  ``` bash
+  GRUB_CMDLINE_LINUX="iomem=relaxed"`
+  ```
+  And regenerate grub config file with `sudo update-grub2` or
+  `sudo grub-mkconfig -o /boot/grub/grub.cfg`. Reboot and then try flashing
+  again.
+
+If above does not resolve the problem the kernel may be compiled with strict
+devmem which prohibits accessing the IOMEM. You should then take different
+Linux system.
```
mikebdp2, line 192: Could suggest using a trusted Linux LiveUSB - maybe even provided by 3mdeb as a bonus to purchase which will already have the configured flashing environment, since the USB flash drives are dirt cheap nowadays _(i.e. https://www.aliexpress.com/item/1005001701059429.html , custom logo possible)_

docs/variants/workstation/installation-manual.md
```diff
@@ -0,0 +1,218 @@
...
+boot:_
+```
+
+You need a workaround to proceed with installation. To boo the installer simply
+type `live-install` and press `ENTER`. It will boot to Ubuntu live and the
+installer will be automatically launched.
+
+Version affected: Dasharo Workstation v0.1
+
+If you see blinking yellow LED and black screen after reboot:
```
mikebdp2, line 211: Is this bug possible to fix by some hardware DIY ?

docs/variants/workstation/installation-manual.md
pietrushnic: done

docs/variants/workstation/installation-manual.md
pietrushnic: done

docs/variants/workstation/installation-manual.md
pietrushnic: @mikebdp2 you are right, please note in future fixes you can incorporate yourself. In most situations like this, I completely agree with your opinion. We will waste fewer resources on going back and fiorth.

docs/variants/workstation/installation-manual.md
pietrushnic: @mikebdp2 grammaly applied

docs/variants/workstation/installation-manual.md
pietrushnic: ok, as I wrote I applied grammarly suggestions

docs/variants/workstation/installation-manual.md
pietrushnic: Added

docs/variants/workstation/installation-manual.md
pietrushnic: @mikebdp2 added TBD placeholder, we definitely plan Dasharo Reference OS

docs/variants/workstation/installation-manual.md
pietrushnic: @mikebdp2 there is not need to fix, only v0.1 is affected, v0.2 and future version have that fixed.

docs/variants/workstation/installation-manual.md
```diff
@@ -0,0 +1,231 @@
...
+
+    ![](../../images/service_mode_warn.jpg)
+
+## Installation
+
+You will need root privileges from now on to proceed. So switch to the root user or
+use __*sudo*__ in each command.
+
+1. Install flashrom v1.1 or newer with your distribution's package manager if
+   you don't have it installed yet. If your distro doesn't provide flashrom,
```
mikebdp2, line 68:
```suggestion
   you don't have it installed yet. If your distro doesn't provide flashrom or provides an outdated one,
```

mikebdp2 (approved):

PR #11, opened 2021-04-02
Dasharo Workstation release notes page - initial version

pietrushnic:


pietrushnic:
It was already fixed in different place.

PR #30, opened 2021-11-09
added kgpe_d16 hw-matrix

kl3mdeb:


macpijan:
@miczyg1 Please review hw config

PR #63, opened 2022-01-20
nv41/flashing_external.md: Update external flashing documentation 

mkopec:

### DIFF
--- a/docs/variants/clevo_nv41/flashing_external.md
+++ b/docs/variants/clevo_nv41/flashing_external.md
@@ -17,6 +17,31 @@ Requirements:
 Ensure that the EC is running updated firmware. Follow the steps outlined in
 [EC firmware update](../ec_update)
 
+## Build flashrom
+
+Currently, the latest flashrom release lacks support for Tiger Lake-U flash
+descriptor processing. Because of this, we need to build flashrom from source.
+
+Install build dependencies:
+
+```bash
+# apt install git build-essential debhelper pkg-config libpci-dev libusb-1.0-0-dev libftdi1-dev meson
+```
+
+Obtain source code:
+
+```bash
+$ git clone https://review.coreboot.org/flashrom.git
+$ cd flashrom
+```
+
+Build flashrom:
+
+```bash
+$ make
+$ sudo make install
+```
+
 ## Setup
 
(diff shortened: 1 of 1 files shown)


docs/variants/clevo_nv41/flashing_external.md
```diff
@@ -17,6 +17,31 @@ Requirements:
 Ensure that the EC is running updated firmware. Follow the steps outlined in
 [EC firmware update](../ec_update)
 
+## Build flashrom
+
+Currently, the latest flashrom release lacks support for Tiger Lake-U flash
```
macpijan, line 22: @mkopec We do not neet chipset support for external flashing, roight?

mkopec: We need it for flashrom to recognize the tiger lake flash descriptor, even when flashing externally. Otherwise we can't flash the BIOS region alone

docs/variants/clevo_nv41/flashing_external.md
```diff
@@ -84,8 +109,19 @@ securely.
 Once you have the backups saved, proceed to flashing the chip - replace
 `[path]` with the path to the binary you want to write - e.g. `build/coreboot.rom`
 
+
+If stock firmware or coreboot version v0.5.0 or older is installed
```
macpijan, line 113: It's should be `Dasharo firmware` version I think

docs/variants/clevo_nv41/flashing_external.md
```diff
@@ -84,8 +109,19 @@ securely.
...
+
+```bash
+# flashrom -p ch341a_spi -c GD25B128B/GD25Q128B -w [path] --ifd -i bios
+```
+
+If coreboot version v0.5.0 or higher is already installed, only the
```
macpijan, line 119: It's external flashing, so i'ts not from OS. Isn't the one command with bios region enough for external flashing?

docs/variants/clevo_nv41/flashing_external.md
macpijan: ok

docs/variants/clevo_nv41/flashing_external.md
mkopec: That's true. Removed

PR #71, opened 2022-02-25
Add Tuxedo IBS15 documentation

mkopec:


docs/variants/tuxedo_ibs15/building.md
```diff
@@ -0,0 +1,61 @@
...
+
+*Warning: Do not run `docker` as root. This command should be executed as your
+current user. If you're having trouble running `docker` on your user account,
+follow the steps outlined in [Requirements](#requirements).*
+
+You should be inside the Docker container. Now run following script to build
+the image:
+
+```bash
+$ ./build.sh build
```
miczyg1, line 57: Can't this script enter the container instead?

docs/variants/tuxedo_ibs15/post_install.md
```diff
@@ -0,0 +1,24 @@
...
+to enable full functionality.
+
+## Touchpad hotkey enablement (Linux)
+
+The touchpad hotkey needs extra setup to function correctly under Linux. To
+enable the touchpad hotkey to work under Linux, follow the steps below:
+
+1. Create a file `/etc/udev/hwdb.d/60-keyboard.hwdb` with the following contents:
+   ```
+   evdev:atkbd:dmi:bvn*:bvr*:svnNotebook:pnNS50MU:*
```
miczyg1, line 13: maybe we could contribute it?

mkdocs.yml
```diff
@@ -107,6 +107,12 @@ nav:
             - 'Logo customization': variants/clevo_nv41/custom_logo.md
             - 'Openness score': variants/clevo_nv41/openness_analysis.md
             - 'LVFS report': variants/clevo_nv41/lvfs_report.md
+        - 'Tuxedo IBS15':
+            - 'Releases': variants/tuxedo_ibs15/releases.md
+            - 'Building': variants/tuxedo_ibs15/building.md
```
miczyg1, line 112: We should unify the section names as in the issue https://github.com/Dasharo/dasharo-issues/issues/37

miczyg1 (changes requested):

docs/variants/tuxedo_ibs15/post_install.md
mkopec: We certainly can. Such a change would go to the systemd repo like so: https://github.com/systemd/systemd/pull/17774

mkdocs.yml
mkopec: Renamed the doc to `Installation`

docs/variants/tuxedo_ibs15/building.md
mkopec: Done, see: https://github.com/Dasharo/coreboot/pull/107/commits/f5438155d9c312f57a4b1d1ed4b534977dbf4dc8

mkopec: Updated docs accordingly.

docs/variants/tuxedo_ibs15/custom_logo.md
```diff
@@ -0,0 +1,32 @@
+# Boot logo replacement instructions
```
pietrushnic, line 1: Why we breaking DRY rule? This site is duplication of https://docs.dasharo.com/variants/clevo_nv41/custom_logo/
When we programming do we copy function everywhere it is needed? Or we create library that can be linked as needed?

docs/variants/tuxedo_ibs15/installation.md
```diff
@@ -0,0 +1,61 @@
+# Dasharo for Tuxedo IBS15 Gen6 - Installation manual
```
pietrushnic, line 1: Another very generic content which most of the time would be reused.

docs/variants/tuxedo_ibs15/post_install.md
```diff
@@ -0,0 +1,24 @@
+# Post-installation setup
```
pietrushnic, line 1: Is this specific to Tuxedo?

mkdocs.yml
```diff
@@ -107,6 +107,12 @@ nav:
             - 'Logo customization': variants/clevo_nv41/custom_logo.md
             - 'Openness score': variants/clevo_nv41/openness_analysis.md
             - 'LVFS report': variants/clevo_nv41/lvfs_report.md
+        - 'Tuxedo IBS15':
```
pietrushnic, line 110: @mkopec I would like to understand why Tuxedo deserve section but Novacustom NV41 is under Clevo NV41. Please be consistent in documentation when you apply some rule, best way to enforce that on others is to add guideline how to approach creation of new section. Last resort all that stuff should be scripted.

docs/variants/tuxedo_ibs15/post_install.md
mkopec: It's common for Clevo laptops with the touchpad disable hotkey.

mkdocs.yml
pietrushnic: Yes, please rename code branches - in light of having potentially multiple OEMs based on Clevo we should improve structure of code to handle that case. @macpijan cc

I know that Felix S. prepared code structure in coreboot for above case, but really don't know it works in practice.

macpijan: > I would like to understand why Tuxedo deserve section but Novacustom NV41 is under Clevo NV41. Please be consistent in documentation when you apply some rule

This is evolving process. With each new machine and project we see more cases to cover. For the NS50 we can see we should use NovaCustom / Tuxedo, and we should probably go back and to the same for NV40. But that would mean we also need to make `novacustom/nv41` mainboard based on the `clevo/nv41`. Wchich can be considered redundant by some, if there is only one such board supported.

> Yes, please rename code branches - in light of having potentially multiple OEMs based on Clevo we should improve structure of code to handle that case.
> I know that Felix S. prepared code structure in coreboot for above case, but really don't know it works in practice.

We already  do that for the NS50 release. We may also change the mainbaord and brancing for the past NV41 but it needs to me done carefuly so we do not break anything.

pietrushnic: Understood, just please optimize layout whenever you can to avoid duplication and incorrect labeling. Please note this repo is part of our customers marketing.

docs/variants/tuxedo_ibs15/building.md
```diff
@@ -0,0 +1,52 @@
+# Dasharo for Tuxedo IBS15 Gen6 - Building manual
```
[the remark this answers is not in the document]
mkopec: Renamed both to `Building manual` as per suggestion.

docs/variants/tuxedo_ibs15/custom_logo.md
mkopec: Moved custom logo instructions to one common location for all platforms.

docs/variants/tuxedo_ibs15/installation.md
mkopec: Most of it is often reusable, yes. However, each device may have some specific caveats that affect it and only it. For example:

- Command for flashing will change after an update enables vboot
- TGL machines need an unreleased (for now) flashrom version

this is why i thought having a separate installation guide per device is a good idea.

docs/variants/tuxedo_ibs15/post_install.md
miczyg1: Let's try to make it happen. @macpijan cc

docs/variants/tuxedo_ibs15/installation.md
pietrushnic: There would be additional effort fort unification. We should maximize use of common content and minimize customized one. Also we should only use customized content when it really is needed not just in case it will be needed in future.

I'm resolving that discussion, but guts tell me not to do that. I hope I will not have to get back to the same rethorics.

docs/variants/tuxedo_ibs15/post_install.md
pietrushnic: We should think how to organize this content in long run, since we now agree to take documentation debt, which we will have to pay with margin later.

pietrushnic (approved):

PR #82, opened 2022-03-02
pre-commit: add basic hooks

macpijan:
Signed-off-by: Maciej Pijanowski <maciej.pijanowski@3mdeb.com>


macpijan:
@pietrushnic There are no problems with these basic hooks, so I will merge it, so they can already be used.

PR #110, opened 2022-03-18
Fixes from pre-commit

Pokisiekk:
Signed-off-by: Kamil Pokornicki kamilpokisiek@3mdeb.com


sulewskiprzemyslaw (approved):

macpijan:
@sulewskiprzemyslaw @Pokisiekk I have added the tags we need to allowed tags in .markdownlint.yaml.

In terms of the line length. Most of these can be easily fixed as shown here: https://github.com/Dasharo/docs/pull/110/commits/f0ac124e7cfbff5e9fdb06edb7bae0dfa791f913
without breaking rendering. So I believe we should go for that. 

@mkopec We should also include this change in our template.

pietrushnic:
I assume here we aggregate all fixes for pre-commit. Please check my changes and fetch frequently since I'm working on those improvements.

pietrushnic:
@macpijan I wonder what should be limit of this PR, since it definitely will be very boring to review.

mkopec:
> @sulewskiprzemyslaw @Pokisiekk I have added the tags we need to allowed tags in .markdownlint.yaml.
> 
> In terms of the line length. Most of these can be easily fixed as shown here: [f0ac124](https://github.com/Dasharo/docs/commit/f0ac124e7cfbff5e9fdb06edb7bae0dfa791f913) without breaking rendering. So I believe we should go for that.
> 
> @mkopec We should also include this change in our template.

@macpijan This doesn't work when we put multiple buttons in one row, then it gets rendered like so: 
image

sulewskiprzemyslaw (approved):

sulewskiprzemyslaw:
@mkopec @pietrushnic @Pokisiekk @macpijan I made formatting changes in release files for `Subscribe to newsletter` buttons. Also, I ran pre-commit again. It no longer showed any formatting errors.

sulewskiprzemyslaw:
@mkopec @macpijan @Pokisiekk @pietrushnic 

I made the last corrections. Now, pre-commit does not show any bugs. Also `mkdocs serve` does not show any warnings related to the internal / external links.

PR #112, opened 2022-03-22
docs/variants/tuxedo_ibs15: add supported IME version

mkopec:
FIXME: Add an actual e-mail address for contact for the ME update


miczyg1:
This is not the way we should go and definitely not the way Dasharo solves things. We should fix it in coreboot, not via IME

PR #118, opened 2022-03-23
unified-test-documentation/dasharo-security/200-tpm-support: Fix comm…

IgorBagnucki:
…and to verify tpm presence

Signed-off-by: Igor Bagnucki <igor.bagnucki@3mdeb.com>

### DIFF
--- a/docs/unified-test-documentation/dasharo-security/200-tpm-support.md
+++ b/docs/unified-test-documentation/dasharo-security/200-tpm-support.md
@@ -31,15 +31,29 @@ be accessed from the operating system.
 1. Power on the DUT.
 1. Boot into the system.
 1. Log into system by using the proper login and password.
-1. If your device is equipped with TPM2.0 - execute the following command
-    in terminal:
+1. Check the version of installed tpm2-tools - execute the following command
+in the terminal:
 
 ```bash
-sudo tpm2_pcrread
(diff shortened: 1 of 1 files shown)


docs/unified-test-documentation/dasharo-security/200-tpm-support.md
```diff
@@ -35,7 +35,7 @@ be accessed from the operating system.
     in terminal:
 
 ```bash
-tpm2_pcrlist
+sudo tpm2_pcrread
```
macpijan, line 38: @IgorBagnucki We know for a fact that tpm2-tools change the names or options of the utilities quite often. I think you found out which was the version introducing this change? Maybe you could add a step with checking version of the tools. And then provide two opsions: tpm2_pcrlist for version > X, tpm2_pcrread for version <= X ?

sulewskiprzemyslaw:
@macpijan @IgorBagnucki  I rewrote the test case and add:

1. Command for checking the version of the tpm2_tools package.
2. Two test options: what to do if tpm2_tools version is 4.0 or higher and what to do if not.

According to the `tpm2-tools` changelog, this command has been re-named in version 4.0 (https://github.com/tpm2-software/tpm2-tools/blob/4.0/CHANGELOG.md).

PR #119, opened 2022-03-23
docs/variants/clevo_nv41/releases.md: add v1.1.0 changelog

mkopec:

### DIFF
--- a/docs/variants/clevo_nv41/flashing_internal.md
+++ b/docs/variants/clevo_nv41/flashing_internal.md
@@ -50,18 +50,18 @@ following command:
 To flash coreboot to the laptop, execute the following command - replace [path]
 with the path to the coreboot image you want to flash, e.g. `build/coreboot.rom`.
 
-If stock firmware or coreboot version below v0.5.0 is currently installed:
+If stock firmware or coreboot version below v1.1.0 is currently installed:
 
 ```bash
 # flashrom -p internal -w [path] --ifd -i bios
 ```
 
-If version v0.5.0 or higher is currently installed, only the RW_SECTION_A
-partition of the flash can be flashed from the operating system. Flash it using
-the following command:
+If version v1.1.0 or higher is currently installed, only the RW_SECTION_A
+partition of the flash needs to be updated. Flash it using the following
+command:
 
 ```bash
 # flashrom -p internal -w [path] --fmap -i RW_SECTION_A
 ```
 
-This command also preserves UEFI settings and the boot order.
+This command also preserves the bootsplash, UEFI settings and the boot order.
--- a/docs/variants/clevo_nv41/flashing_internal.md
+++ b/docs/variants/clevo_nv41/flashing_internal.md
@@ -1,8 +1,8 @@
-# coreboot internal flashing (Ubuntu 20.04)
+# Dasharo internal flashing (Ubuntu 20.04)
 
 ## Intro
 
-Flashing coreboot can be done from Linux using flashrom with the internal
+Flashing Dasharo can be done from Linux using flashrom with the internal
 programmer. This document describes the process of building, installing and
 running flashrom on Ubuntu 20.04.
 
@@ -45,12 +45,12 @@ following command:
 # flashrom -p internal -r dump.rom
 ```
 
-## Flashing coreboot
+## Flashing Dasharo
 
-To flash coreboot to the laptop, execute the following command - replace [path]
-with the path to the coreboot image you want to flash, e.g. `build/coreboot.rom`.
+To flash Dasharo to the laptop, execute the following command - replace [path]
+with the path to the Dasharo image you want to flash, e.g. `build/coreboot.rom`.
 
-If stock firmware or coreboot version below v1.1.0 is currently installed:
+If stock firmware or Dasharo version below v1.1.0 is currently installed:
 
 ```bash
 # flashrom -p internal -w [path] --ifd -i bios
--- a/docs/variants/clevo_nv41/releases.md
+++ b/docs/variants/clevo_nv41/releases.md
@@ -8,6 +8,35 @@ For details about our release process please read
 
 [Subscribe to Clevo NV41MZ Dasharo Release Newsletter](https://newsletter.3mdeb.com/subscription/T61MyO2sP){ .md-button .md-button--primary .center }
 
+## v1.1.0 - 2022-03-23
+
+### Added
+
+- Add Dasharo Tools Suite network boot integration
+- Add a persistent bootlogo implementation
+
+### Known issues
+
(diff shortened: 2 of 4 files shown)


docs/variants/clevo_nv41/releases.md
```diff
@@ -8,6 +8,35 @@ For details about our release process please read
...
+- [Windows 10 SD card reader driver needs manual installation on NV41MB](https://gitlab.com/novacustom/dasharo-compatibility/-/issues/16)
+- [Low level interfering crackling/popping of the speakers while playing no sound](https://gitlab.com/novacustom/dasharo-compatibility/-/issues/13)
+- [Laptop cannot output video via the Tunderbolt 4 USB Type-C port](https://gitlab.com/novacustom/dasharo-compatibility/-/issues/7)
+
+
+### Binaries
+
+[clevo_nv41mz_v1.1.0.rom](https://3mdeb.com/open-source-firmware/Dasharo/clevo_nv41mz/clevo_nv41mz_v1.1.0.rom){ .md-button }
+[sha256](https://3mdeb.com/open-source-firmware/Dasharo/clevo_nv41mz/clevo_nv41mz_v1.1.0.rom.sha256){ .md-button }
+[sha256.sig](https://3mdeb.com/open-source-firmware/Dasharo/clevo_nv41mz/clevo_nv41mz_v1.1.0.rom.sha256.sig){ .md-button }
```
pietrushnic, line 31: @mkopec doesn't pre-commit complain on those long lines? I noted that if you have bullet list then pre-commit don't care about line length otherwise there are tricks need to avoid problems and complains with this lines. Of course because this is button it probably cannot be in bullet list.

This is how I [resolved that in fix_branch](https://github.com/Dasharo/docs/pull/110/commits/98d4a7f93fbb1056344eef8789ef0802a0bfd279). I'm not sure if this is acceptable strategy so please take it with grain of salt. @macpijan please let me know also what is your view on that.

pietrushnic (changes requested):
One small question.

docs/variants/clevo_nv41/releases.md
pietrushnic: @macpijan I also have similar feeling that it may be hard to maintain, alternatively we can add some short labels, we just need some pattern. I see no need for re-enumeration after adding new release, numbers are just labels which can increase forever.

I don't know if maintenance of that would be expensive, we doing that just once and we should never change that. What is the reason to change links/labels for some old releases?

Proposed fix, not talking about tags, will not work for following reasons:

1. [This](https://github.com/Dasharo/docs/commit/f0ac124e7cfbff5e9fdb06edb7bae0dfa791f913#diff-b0f1b7a0ce7eb3de7caeb1c99af3af7c4c4944b7c4475e94d295f2522500efd6R60) is already done in many palaces but link still is longer then 80 characters.
2. [This](https://github.com/Dasharo/docs/commit/f0ac124e7cfbff5e9fdb06edb7bae0dfa791f913#diff-b0f1b7a0ce7eb3de7caeb1c99af3af7c4c4944b7c4475e94d295f2522500efd6R9) will not work for Talos II because name is too long. I proposed change of button label which always would be shorter [here](https://github.com/Dasharo/docs/pull/110/commits/98d4a7f93fbb1056344eef8789ef0802a0bfd279#diff-342c0ec9d784cffe1926a64dfd8e7690953a4fdc1a618216eb3d9b051409f24fR9)

From what I understand tags may be good solution, but reasonable labeling could also work. If we talking about labels there is standard for that in CommonMark which is used by pre-commit [here](https://spec.commonmark.org/0.30/#link-reference-definitions), but if we agree about using tags then let's do it.

mkopec: @macpijan As I mentioned in https://github.com/Dasharo/docs/pull/110#issuecomment-1076510267, the solution proposed in https://github.com/Dasharo/docs/pull/110#issuecomment-1074884316 is not perfect

mkopec:
I just realized, we changed the FMAP layout in this release, and the RW_SECTION_A region was moved and resized, this means we need to update the update procedure to flash the whole BIOS region instead

docs/variants/clevo_nv41/releases.md
macpijan: Maybe let's try with the tags then. It would be nice if they can follow some rule (e.g. component_version) so they can be automatically generated.

mkopec: @macpijan The tags can be named and placed anywhere in the doc it seems, so i.e. this will render correctly:

```
[clevo_nv41mz_v1.1.0.rom][v1.1.0_rom]{ .md-button }
[sha256][v1.1.0_hash]{ .md-button }
[sha256.sig][v1.1.0_sig]{ .md-button }

[v1.1.0_rom]: https://3mdeb.com/open-source-firmware/Dasharo/clevo_nv41mz/clevo_nv41mz_v1.1.0.rom
[v1.1.0_hash]: https://3mdeb.com/open-source-firmware/Dasharo/clevo_nv41mz/clevo_nv41mz_v1.1.0.rom.sha256
[v1.1.0_sig]: https://3mdeb.com/open-source-firmware/Dasharo/clevo_nv41mz/clevo_nv41mz_v1.1.0.rom.sha256.sig
```

I'll update the entire release page and check if this satisfies the linter

pietrushnic: @mkopec @macpijan I like this solution. @mkopec would you mind to fix that in #110 ? With new changes I will follow that rule, maybe we should also write that somewhere or enforce in pre-commit?

mkopec: @pietrushnic Sure, I'll fix it in #110 too

docs/variants/clevo_nv41/flashing_internal.md
```diff
@@ -50,18 +50,18 @@ following command:
 To flash coreboot to the laptop, execute the following command - replace [path]
 with the path to the coreboot image you want to flash, e.g. `build/coreboot.rom`.
 
-If stock firmware or coreboot version below v0.5.0 is currently installed:
+If stock firmware or coreboot version below v1.1.0 is currently installed:
```
macpijan, line 53: @mkopec Maybe it should be `Dasharo version` ?

mkopec: Yes. Changed most occurences of coreboot to Dasharo

docs/variants/clevo_nv41/releases.md
mkopec: Applied to nv41 in #110 too

mkopec: @pietrushnic Applied to Talos and KGPE-D16 as well

PR #120, opened 2022-03-23
clevo_nv41/fwupd_usage: Disable secure boot for fwupd

IgorBagnucki:
Signed-off-by: Igor Bagnucki <igor.bagnucki@3mdeb.com>

### DIFF
--- a/docs/variants/clevo_nv41/fwupd_usage.md
+++ b/docs/variants/clevo_nv41/fwupd_usage.md
@@ -120,7 +120,9 @@ $ sudo ninja -C build install
 
 ## Updating firmware from LVFS
 
-1. Use the following command to update firmware with fwupd:
+1. Disable Secure Boot.
+
+2. Use the following command to update firmware with fwupd:
 
     *Note: You may be asked which device to update. If you see a prompt, select
     NV4XMB,ME,MZ*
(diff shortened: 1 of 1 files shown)


mkopec (approved):

docs/variants/clevo_nv41/fwupd_usage.md
```diff
@@ -120,7 +120,9 @@ $ sudo ninja -C build install
 
 ## Updating firmware from LVFS
 
-1. Use the following command to update firmware with fwupd:
+1. Disable Secure Boot.
```
macpijan, line 123: @IgorBagnucki Can we elaborate how to do that? Or link to a document where we already have explained that?

macpijan: Also, the lockdown mode may be a bit different depending on a distro used.

How about also disabling lockdown as suggested here: https://github.com/flashrom/flashrom/issues/59 ?

mkopec: We don't have a specific procedure for disabling Secure Boot, closest we have is https://docs.dasharo.com/unified-test-documentation/dasharo-security/206-secure-boot/ which describes *enabling* Secure Boot, but the process is almost identical.

For disabling kernel lockdown, as far as i can tell the sysrq+x disable was never accepted upstream and only some kernels had it for some time, but dropped it due to security problems.

macpijan: @IgorBagnucki Let's link it then with the releveant description and then we can merge

IgorBagnucki: @macpijan The link is added.

PR #126, opened 2022-03-25
Docker fwdump doc

PrzemyslawBanasiak13:
Signed off by Przemyslaw Banasiak [przemyslaw.banasiak@3mdeb.com]


mkopec (approved):

macpijan:
This is not linked anywhere, so probably it's not very useful. Also, we want to replace it with DTS as soon as possible.

PR #136, opened 2022-04-04
Draft: add initial release process documentation

pietrushnic:
Signed-off-by: Piotr Król <piotr.krol@3mdeb.com>

### DIFF
--- a/docs/dev-proc/standard-release-process.md
+++ b/docs/dev-proc/standard-release-process.md
@@ -1,13 +1,38 @@
 # Standard Release Process
 
 Following procedure is generic description of release process of firmware for
-supported hardware platforms. Precise steps and any difference from standard
-process are described in platform specific documentation.
+supported hardware platforms and targets open-source firmware developers.
+Precise steps and any difference from standard process are described in
+platform specific documentation.
 
-Description here is, intentionally, Open Source Firmware framework agnostics
+Description here is, intentionally, open-source firmware framework agnostics
 and should be maintained in that way.
 
-## Process steps
+## Initial release
+
+For platforms not supported by Dasharo there is need for initial release, which
+consist of [Dasharo Common
+Base](../../osf-trivia-list/dasharo/#what-is-dasharo-common-base). Initial
+release should have established test scope defined in platform [test
+matrix](../../unified-test-documentation/overview/#test-matrix) section.
+
+1. Fetch upstream project
+1. Checkout branch `<platfrom>/release`
+    - `coreboot` - use most recent commit on upstream `master`
+    - `edk2` - use most recent stable tag
+1. Create branch `<platform>/rel_v0.1.0`
+1. Cherry-pick required patches from `dasharo-common-base` branch
+1. Apply all other modifications necessary to  provide initial support for
+   `<platform>` according to platform test matrix scope.
+1. Run platform regression test suite
+1. If results are accepted merge it to `<platform>/release` branch
+1. Add tag, which should trigger CI and publish binaries
+1. Merge release branch to develop
+
+## Regular release
+
+Following procedure applies to platform for which initial release was already
+published.
 
 1. Checkout new branch `<platform>/rel_vX.Y.Z` from recent commit on
    `<platfrom>/release` - to understand versioning scheme please read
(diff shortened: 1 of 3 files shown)


docs/osf-trivia-list/dasharo.md
```diff
@@ -21,3 +21,9 @@ it may consist of:
 * Fixes introducing Dasharo SMBIOS/DMI tables format according to [Dasharo
   Product Guidelines](../../dev-proc/smbios-rules)
 * Fixes for submodules building
+
+Those improvements are stored on `dasharo/common-base` branch in relevant
```
mkopec, line 25: Other names I was considering: `dasharo/dasharo`, `dasharo/base`

docs/dev-proc/standard-release-process.md
```diff
@@ -1,13 +1,38 @@
...
 
 Following procedure is generic description of release process of firmware for
-supported hardware platforms. Precise steps and any difference from standard
-process are described in platform specific documentation.
+supported hardware platforms and targets open-source firmware developers.
+Precise steps and any difference from standard process are described in
+platform specific documentation.
 
-Description here is, intentionally, Open Source Firmware framework agnostics
+Description here is, intentionally, open-source firmware framework agnostics
```
mkopec, line 8:
```suggestion
Description here is, intentionally, open-source firmware framework agnostic
```

mkopec, line 8:
```suggestion
Description here is, intentionally, open-source firmware framework agnostic
```

docs/dev-proc/standard-release-process.md
```diff
@@ -1,13 +1,38 @@
...
+
+1. Fetch upstream project
+1. Checkout branch `<platfrom>/release`
+    - `coreboot` - use most recent commit on upstream `master`
+    - `edk2` - use most recent stable tag
+1. Create branch `<platform>/rel_v0.1.0`
+1. Cherry-pick required patches from `dasharo-common-base` branch
+1. Apply all other modifications necessary to  provide initial support for
+   `<platform>` according to platform test matrix scope.
+1. Run platform regression test suite
```
miczyg1, line 27: The steps below assume that platform port is ready. No regression would be run without a port that is somewhat bootable

docs/osf-trivia-list/dasharo.md
miczyg1: Another proposal:

A set of patch files applicable on the recent stable coreboot tag (from which we should base all new ports, always a stable tag as a base for new port, Dasharo release). They would be stored on e.g. dasharo/dasharo repository. coreboot seems not to be a good place as it would introduce another perm branch.

These patches could conform to release cycles, i.e. a new release of patches would occur on a new coreboot upstream release quarterly.

docs/dev-proc/standard-release-process.md
pietrushnic: @miczyg1 what change you expect in procedure? Previous point explain that all modifications to provide initial support should be applied to the branch.

docs/osf-trivia-list/dasharo.md
macpijan: > Are you sure we should base on stable coreboot tag? I thought we had that discussion, and we decided that there is too much going on in coreboot to base on "stable" branch, which has no guarantees. IMO there is no logical justification of that statement instead of having upstream-managed known point in time. There is no other value in stable IMO.

I'm quite surprised to hear that, as my impression was that you always opted in for the tag, rather than some arbitrary commit.

macpijan: If we want to name it `Dasharo Common Base` then `dasharo/common-base` branch totally makes sense to me.

pietrushnic: @macpijan yes I did opted for for tag until I realized there is no way of changing mindset in coreboot leadership and community. Tags in coreboot are and will be just arbitrary point in time, so if we choose any other arbitrary point the value is the same, we just have smaller difference in code to take care what should mean easier upstream. Anything we need in Dasharo Common Base should be considered for upstream e.g. DMI settings could be just Kconfig options.

What we can think in future is having wide quality tests across Dasharo targets which can claim about quality of coreboot, if there will be any difference. Then based on test results for Dasharo targets we could judge if there is any place in the tree which is better than the other, but this is long story which means establishing quality control for huge project.

PR #162, opened 2022-04-22
docs/common-coreboot-docs/fwupd_usage.md: change firmware update command

Pokisiekk:


macpijan:
@mkopec What should we to with this MR?

mkopec:
> @mkopec What should we to with this MR?

fwupdmgr should work just as well, so changing the command shouldn't be needed. Could be that updating from the embargo repo, where packages aren't signed with the fwupd key and fwupdmgr rejects them, requires this workaround.

sulewskiprzemyslaw:
@Pokisiekk In relation to @mkopec writes about I decided to close this PR

PR #181, opened 2022-05-11
dasharo-performance/401-cpu-temperature.md: add file

Pokisiekk:

### DIFF
--- a/docs/osf-trivia-list/me.md
+++ b/docs/osf-trivia-list/me.md
@@ -1,4 +1,4 @@
-## Introduction
+# Introduction
 
 Until Nehalem (ME version 6) it was possible to remove ME firmware completely.
 Starting with Nehalem, ME firmware can't be completely removed.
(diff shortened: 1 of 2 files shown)


sulewskiprzemyslaw:
@Pokisiekk I wonder if we should use package `lm-sensors` instead of reading the temperature form file. I think that it would be better because in that case we also have information about which core has which temperature and what are the defined values for high temperature and critical temperature.

Example output from lm-sensors:
`acpitz-acpi-0
Adapter: ACPI interface
temp1:        +27.8°C  (crit = +106.0°C)
temp2:        +29.8°C  (crit = +106.0°C)

coretemp-isa-0000
Adapter: ISA adapter
Package id 0:  +42.0°C  (high = +85.0°C, crit = +105.0°C)
Core 0:        +38.0°C  (high = +85.0°C, crit = +105.0°C)
Core 1:        +40.0°C  (high = +85.0°C, crit = +105.0°C)
Core 2:        +42.0°C  (high = +85.0°C, crit = +105.0°C)
Core 3:        +40.0°C  (high = +85.0°C, crit = +105.0°C)
`

sulewskiprzemyslaw (approved):

PR #192, opened 2022-05-20
add links and tables with test results in releases.md files

Pokisiekk:


sulewskiprzemyslaw:
@Pokisiekk I talked with @miczyg1. For this time we only add links to the test results spreadsheets - files `releases.md` are generated automatically so it would be better to improve the generator than add test results manually.

PR #198, opened 2022-05-25
variants/{novacustom_ns5x,tuxedo_ibs15}: flash the COREBOOT region

mkopec:
Vboot was disabled on these laptops, so the flashing instructions need to be updated.

### DIFF
--- a/docs/variants/novacustom_ns5x/installation.md
+++ b/docs/variants/novacustom_ns5x/installation.md
@@ -77,11 +77,17 @@ Steps for initial Dasharo installation:
 
 ### Updating Dasharo
 
-If Dasharo is currently installed, only the RW_SECTION_A partition of the flash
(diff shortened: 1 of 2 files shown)


mkopec:
@macpijan ping

macpijan (changes requested):
If the procedure is different depending on the version, maybe we should have both methods and indicate which version support which method.

macpijan:
@mkopec

mkopec:
@macpijan Updated. For NVC NS5x, Dasharo v1.1.0 and newer has Vboot disabled. Tuxedo IBS15 without Vboot wasn't released yet, so for now I reverted the command  back to `--ifd -i bios`

macpijan (approved):

macpijan:
picked in: https://github.com/Dasharo/docs/pull/333

PR #212, opened 2022-06-03
docs/common-coreboot-docs/dasharo_tools_suite.md: v0.3.0

macpijan:
Signed-off-by: Maciej Pijanowski <maciej.pijanowski@3mdeb.com>


docs/common-coreboot-docs/dasharo_tools_suite.md
```diff
@@ -2,30 +2,89 @@
...
-## Launching DTS
+* bootable USB stick image.
+* bootable over network (iPXE),
+
+## Flavors
+
+There is a common base, but there might be multiple flavors of the DTS images.
+Currently, there are:
+
+* CE
```
miczyg1, line 23:
```suggestion
* CE - Community Edition
```

docs/common-coreboot-docs/dasharo_tools_suite.md
```diff
@@ -2,30 +2,89 @@
...
+      information
+    - flashrom, fwupd, and many more useful tools
+    - can drop to shell to update the firmware manually
+* OEM
+    - on top of that, provides tools for automatic firmware deployment and
+      rollback (switching to Dasharo back and forth)
+
+## Bootable USB stick
+
+This documentation is compatible with the `v0.3.0` version of the DTS.
```
miczyg1, line 34: This sounds like a problem. It would mean we need a versioning for the documentation. Now:
1. Does it mean the documentation for newer versions will not be backwards compatible?
2. How to check the documentation for older version in the future, e.g. when we release v1.2.3 for example, how to check v1.1.1 if someone still uses it?

We could maybe assume that the documentation is extend-only where each extension is marked with the first version since feature introduction. I.e. supported since version vx.y.z?

docs/common-coreboot-docs/dasharo_tools_suite.md
```diff
@@ -2,30 +2,89 @@
...
+* [DTS CE v0.3.0 downloaded](https://cloud.3mdeb.com/index.php/s/Q7pAppm4gnRCef9/download)
+
+### Launching DTS
+
+* Flash the downloaded `dts-base-image-ce-v0.3.0.wic.gz` image onto USB stick.
+  - you can use cross-platform GUI installer - [Etcher](https://www.balena.io/etcher/)
+  - you can also use `dd` to flash from command line
+
+```bash
+gzip -cdk dts-base-image-ce-v0.3.0.wic.gz | sudo dd of=/dev/sdX bs=16M status=progress
```
miczyg1, line 50: Using bs of 16MB will certainly cause some post-dd sync of the data on the stick. If someone pulls out the stick right after dd, data corruption is guaranteed. The typical MTU for USB is 512 bytes. I would advise adding a next step which will instruct to safely eject the drive from the system using the desktop manager and wait until the remaining writes actually occur.

docs/common-coreboot-docs/dasharo_tools_suite.md
```diff
@@ -2,30 +2,89 @@
...
+
+* A menu should appear
+
+* Enter `1` to dump logs with hardware from your device
+
+![](../images/dts-hcl-run.png)
+
+## Bootable over network
+
+This documentation is compatible with the `v0.1.0` version of the DTS.
```
miczyg1, line 67: There we go. Some parts of the document will be compatible with version X and some parts with version Y possibly causing confusion. Now I also think that it may be wise to show the most common usage here and link to more detailed documentation in the DTS repository. Documentation on DTS repository would be tagged along with the code and get rid of the problem of the versioned documentation here.

docs/common-coreboot-docs/dasharo_tools_suite.md
mkopec, line 50: `conv=fdatasync` should handle this for us:

```suggestion
gzip -cdk dts-base-image-ce-v0.3.0.wic.gz | sudo dd of=/dev/sdX bs=16M status=progress conv=fdatasync
```

mkopec: `conv=fdatasync` should handle this for us:

```suggestion
gzip -cdk dts-base-image-ce-v0.3.0.wic.gz | sudo dd of=/dev/sdX bs=16M status=progress conv=fdatasync
```

docs/common-coreboot-docs/dasharo_tools_suite.md
macpijan: > Does it mean the documentation for newer versions will not be backwards compatible?

I don't know maybe. I can easily image that yes. Everything can change here at this point.

docs/common-coreboot-docs/dasharo_tools_suite.md
macpijan: I left it this way, becasue we have only two varianst in the wild at the moment:
- v0.1.0 - deployed on boot.3mdeb.com
- v0.3.0 - USB stick

With the next release, we can create a separate DTS documentation, which will be versioned for a given release.

PR #325, opened 2022-10-05
docs/variants/novacustom_ns5x_7x/nv4x/post_install.md: add post-install step on Windows 11

Pokisiekk:


macpijan (commented):
For GUI guides, having screenshots would be better. But it's already much better than nothing, so let's start with it.

PR #379, opened 2022-11-14
docs/unified/clevo/post-install.md: add nv4x adl verbs fix

mkopec:

### DIFF
--- a/docs/unified/clevo/post-install.md
+++ b/docs/unified/clevo/post-install.md
@@ -129,23 +129,8 @@ The headset jack in NV4x 12th Gen (Alder Lake) needs a fix that is available
 starting with kernel version v6.0. If you are using an older kernel, you need
 to add the fix manually:
 
-1. Create a file `/lib/firmware/hda-init.fw` with the following contents:
-
     ```bash
-    [codec]
-    0x10ec0256 0x15584041 0
-
-    [model]
-    headset-mode-no-hp-mic
-
-    [pincfg]
-    0x19 0x03a11120
-    ```
-
-1. Create a file `/etc/modprobe.d/audio-jack.conf` with the following contents:
-
-    ```bash
-    options snd_hda_intel patch=hda-init.fw
+    curl -sSf https://raw.githubusercontent.com/Dasharo/dasharo-tools/main/clevo/nv4x-audio-fixup | sudo sh
     ```
 
 1. Reboot. The audio jack will now work correctly.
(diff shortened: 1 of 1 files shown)


docs/unified/clevo/post-install.md
```diff
@@ -122,3 +122,30 @@ power saving tweaks to enable sleep mode while a SATA disk is installed.
 
 1. Download the script: [link](https://raw.githubusercontent.com/Dasharo/dasharo-tools/main/clevo/sata-suspend-fixup.bat)
 2. Double click on the script to install the tweak
+
+### Headset jack fix for NV4x 12th Gen
+
+The headset jack in NV4x 12th Gen (Alder Lake) needs a fix that is available
+starting with kernel version v6.0. If you are using an older kernel, you need
```
macpijan, line 129: We can link to this commit/patch.

docs/unified/clevo/post-install.md
```diff
@@ -122,3 +122,30 @@ power saving tweaks to enable sleep mode while a SATA disk is installed.
...
+to add the fix manually:
+
+1. Create a file `/lib/firmware/hda-init.fw` with the following contents:
+
+    ```bash
```
macpijan, line 134: Don't we prefer using scripts https://github.com/Dasharo/dasharo-tools/tree/main/clevo to create these automatically to simplify post-install steps?
We can imagine we will gather more and more scripts like this and provide a single
post-install script in the documentation, which will detect what fixups are needed
on the given OS and given kernel/device/firmware/etc.

macpijan:
We need confirmation from @Pokisiekk or @sulewskiprzemyslaw that this in fact fixes the problem with headset.

docs/unified/clevo/post-install.md
mkopec: Changed the doc to use the script instead of creating files manually.

docs/unified/clevo/post-install.md
mkopec: Added a link to the specific commit

macpijan (commented):
LGTM. Still need approve from QA

pietrushnic:
@macpijan @mkopec post-install stuff should be handled by fwupd/LVFS later IIUC?

macpijan:
I do not think so

mkopec:
Not sure if it's possible or even desirable - this is more of a first-setup thing during initial Dasharo deployment and/or OS installation

PR #400, opened 2022-12-02
Test-matrix and hardware-matrix Novacustom refactor

Pokisiekk:


docs/variants/novacustom_nv4x_tgl/test-matrix.md
```diff
@@ -1,25 +1,23 @@
-# Test matrix - NovaCustom NV41
+# Test matrix
```
sulewskiprzemyslaw, line 1: @Pokisiekk I think, that we should leave separate test matrixes for two tested models. When creating and maintaining documentation, let's assume that even if T-M differs by one item, it should be separated.

sulewskiprzemyslaw (changes requested):

docs/variants/novacustom_nv4x_tgl/test-matrix.md
Pokisiekk: OK, separated

sulewskiprzemyslaw (approved):

PR #408, opened 2022-12-08
Docs guidelines

TomaszAIR:
Reviewed in internal repos


kotylamichal (approved):
It is the same doc as we review in the internal discussion (except a few formatting corrections), so LGTM

PR #465, opened 2023-02-06
Remove Ubuntu 21.04 reference

nedpfeiffer:
Ubuntu 21.04 is end-of-life. 

Source: https://fridge.ubuntu.com/2022/01/21/ubuntu-21-04-hirsute-hippo-end-of-life-reached-on-january-20-2022/

### DIFF
--- a/docs/variants/msi_z690/building-manual.md
+++ b/docs/variants/msi_z690/building-manual.md
@@ -7,7 +7,7 @@ compatible with MSI PRO Z690-A WIFI DDR4.
 
 ## Requirements
 
-* `Ubuntu 20.04/21.04/22.04` as a host OS was tested
+* `Ubuntu 20.04/22.04` as a host OS was tested
 * Internet connection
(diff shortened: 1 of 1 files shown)


pietrushnic:
@nedpfeiffer thank you for you contribution. I agree test scope should cover only non-EOL OS releases, but I guess the fact we tested on older version may be useful historically to some users. To change test scope we need next release, when we will omit EOL version. For now I would not change above.

@macpijan @sulewskiprzemyslaw WDYT?

sulewskiprzemyslaw:
@pietrushnic I agree with you. We should take this PR into account during preparing the next release.

Maybe we should start preparing dedicated branches and PRs for releases? These branches might contain all information dedicated for release (release notes, minor changes such as this one, etc).

macpijan:
This line simply tells it was tested on this release. The fact that it became OEL in between does not change it.

PR #475, opened 2023-02-14
docs/variants/msi_z690/recovery.md: paragraph order

nedpfeiffer:
Per the Dasharo [documentation guidelines](https://docs.dasharo.com/dev-proc/documentation-guidelines/) the most important information should be at the top of the documentation. Since flashing with the MSI Flash BIOS Button does not currently work, I moved it to the bottom. Let me know what you think.


pietrushnic (approved):

pietrushnic:
@miczyg1 I guess you had opinion about that so I leave it unmerged so you can let us know if you have any concerns about this change.

pietrushnic:
@nedpfeiffer thank you for your contribution.

miczyg1:
Yes, I think MSI FlashBIOS button will be the most important recovery mechanism once we get it working thus I was not sure whether we should move it lower. As long as the button does not work, it does not do any harm. If we manage to get it working, I will move it up again. For now I am fine with merging it.

miczyg1 (approved):

PR #478, opened 2023-02-17
novacustom: fan profiles description

macpijan:


BeataZdunczyk (approved):

docs/unified/novacustom/fan-profiles.md
```diff
@@ -0,0 +1,23 @@
...
+| 100              | 100           |
+
+## Performance fan profile
+
+| Temperature [°C] | Fan speed [%] |
+|------------------|---------------|
+| 0                | 25            |
+| 55               | 35            |
+| 75               | 60            |
+| 100              | 100           |
```
sulewskiprzemyslaw, line 23: @macpijan What is the characteristic between the indicated points? Linear? I think such information would be helpful for users.

sulewskiprzemyslaw (changes requested):

docs/unified/novacustom/fan-profiles.md
mkopec: Values inbetween curve points are linearly interpolated, yes.

For temperatures below the first point the fans will be stoppend and above the last point the fans will be set to 100%.

sulewskiprzemyslaw: So we should add such an information

mkopec: Added a note about this.

sulewskiprzemyslaw (approved):

PR #496, opened 2023-03-06
Dasharo se

miczyg1:


docs/osf-trivia-list/dts.md
```diff
@@ -2,10 +2,13 @@
...
 
-Dasharo Tools Suite Supporters Entrance (DTS SE) is an access to Dasharo
-firmware updates for your hardware. Thanks to Dasharo Tools Suite you can
-execute initial deploy, with SE you could get updates or different build
-configurations.
+Dasharo Tools Suite Supporters Entrance (DTS SE) is an extended version of
+[DTS](../dasharo-tools-suite/overview.md) provided to Dasharo Supporters via
+the [Supporters Entrance subscription](../ways-you-can-help-us.md#become-a-dasharo-supporter).
+DTS is like a swiss army knife for Dasharo firmware: you can execute initial
+deploy, firmware update and even dump machine logs for debugging and
```
pietrushnic, line 9:
```suggestion
deployment, firmware update and even dump machine logs for debugging and
```

docs/osf-trivia-list/dts.md
```diff
@@ -2,10 +2,13 @@
...
+development support. But with SE you could get updates for your own, different
```
pietrushnic, line 10:
```suggestion
development support. But with SE you could get updates for your own, customized
```

docs/ways-you-can-help-us.md
```diff
@@ -66,6 +66,32 @@ open-source firmware mission. It is straightforward if you don't have time for
...
+* or other platforms
+
+As a Dasharo Supporter you will get:
+
+* The latest Dasharo releases issued by Dasharo Team
+* Special Dasharo Updates for Supporters –  Community receives one update per
+  year, but Supporters are not limited to it. Number of updates depends on the
+  number of Dasharo Subscriptions sold and the availability of other funding
+  (e.g., NLNet, corporate sponsors, community donations).
+* Dasharo Primer Support through invite-only Matrix channel
```
pietrushnic, line 91: where we often gather Supporters opinion  and share exciting news about Dasharo direction

pietrushnic (changes requested):
added some minor comments

docs/ways-you-can-help-us.md
miczyg1: Applied https://github.com/Dasharo/docs/pull/496/files#diff-416aa4deddee57bbb844c0d8516e5fe680344bb7a2e77f628d019f4e611223afR91

pietrushnic (approved):

docs/ways-you-can-help-us.md
```diff
@@ -66,6 +66,33 @@ open-source firmware mission. It is straightforward if you don't have time for
...
+Or buy the hardware including the subscriptions from us, e.g.:
+
+* [MSI PRO Z690-A DDR4 motherboard with Dasharo Supporters Entrance](https://3mdeb.com/shop/open-source-hardware/dasharo-compatible-with-msi-pro-z-690a-ddr4-motherboard/)
+* [MSI PRO Z690-A DDR4 with Dasharo Supporters Entrance – Full PC Build](https://3mdeb.com/shop/open-source-hardware/dasharo-compatible-with-msi-pro-z-690a-ddr4-full-pc-build/)
+* or other platforms
+
+As a Dasharo Supporter you will get:
+
+* The latest Dasharo releases issued by Dasharo Team
+* Special Dasharo Updates for Supporters –  Community receives one update per
```
TomaszAIR, line 87: > Community receives one update per year

Are we sure about that? Do we agree that publicly available update will be released one per year?

docs/ways-you-can-help-us.md
```diff
@@ -66,6 +66,33 @@ open-source firmware mission. It is straightforward if you don't have time for
 other contributions, and there are multiple options - feel free to choose one
 that works for you.
 
+### Become a Dasharo Supporter
+
+Buy the Dasharo Supporters Entrance subscription if you own/use/buy a machine
+running Dasharo firmware:
+
+* [1-year subcription](https://3mdeb.com/shop/open-source-hardware/subscription-for-dasharo-compatible-with-msi-pro-z-690a-ddr4/)
```
TomaszAIR, line 74: Should we not have options just for Dasharo SE for 1/2/3 years without `compatible with MSI Pro Z690-A` ? In case there will be more options (like e.g. with stable releases for OptiPlex or T1650).

We had got so many ideas (Dasharo Patreon/ CE/ SE/ OEM/ Subscription) that one might get lost. We should probably pick one and try to stay with it.

TomaszAIR (commented):
Add two questions about these changes

docs/ways-you-can-help-us.md
pietrushnic: Agree, we should generalize that, but key problem is that subscription most likely make sense for given platform price point.

e.g. it may make sense to add subscription of 60EUR for platform that cost 1k EUR, but does it make sense for platform that cost 100-300EUR?

> We had got so many ideas (Dasharo Patreon/ CE/ SE/ OEM/ Subscription) that one might get lost. We should probably pick one and try to stay with it.

Agree, but do we know what will work? We change things because we learn all the time and we hope for continues improvements. I believe we building better options then those before. We should have clear distinction between SE buyers and donors through Patreon membership (let's be honest Patreon names were invented because you are forced to do that when you create Patreon). Some community members thought Patreon would be much better then subscription and we still get such feedback e.g. in PC Engines survey.

To sum up, I'm all about one option because I know we need focus, but @TomaszAIR do you have idea what that would be? What will work?

So maybe at this point we should at least make distinction clear and if community and customers don't like it they will tell us.

docs/ways-you-can-help-us.md
TomaszAIR: I'm asking about this piece of text to make sure what kind of attitude we have towards the update now. The earlier plan was that initial deployment is publicly available and possibility to update from within the menu reserved for Dasharo SE.

I also understand that the last release required that the upgrade capability be made available to all users, and that's why we implemented the first version of the upgrade from the menu - I preferred to elaborate/describe in the comments to make sure we're on-the-same-page here.

docs/ways-you-can-help-us.md
TomaszAIR: > e.g. it may make sense to add subscription of 60EUR for platform that cost 1k EUR, but does it make sense for platform that cost 100-300EUR?

Agree, I did not think about that earlier.

> Some community members thought Patreon would be much better then subscription and we still get such feedback e.g. in PC Engines survey.

I did not know about that aspect. I hope we will get some brainstorming sessions about DTS/Dasharo/subscription model and will be able to present clear news on DUG#1.

docs/ways-you-can-help-us.md
pietrushnic: Right now the idea is precise as described by @miczyg1 [here](https://github.com/Dasharo/docs/pull/497/files). Please let me know is this is clear enough, if not we will try to improve the description.

docs/ways-you-can-help-us.md
pietrushnic: Yes, I also hope for that, but there is very little time for providing clear solution. We slowly getting there. Let's assign Dasharo Subscription product generalization for all supported hardware platforms. Internally we should count how many subscriptions we sold and what do we release we should communicate during DUGs.

docs/ways-you-can-help-us.md
TomaszAIR: Now it is clear. To summarize. We want to release one community release per year that bumps up the minor version and up to two Dasharo SE releases (can be zero, one or two) that bumps up the patch version. The community release will have all the changes of the previously released Dasharo SE releases.

PR #498, opened 2023-03-06
Dasharo se semver

miczyg1:


docs/dev-proc/versioning.md
```diff
@@ -13,6 +13,23 @@ notes. Link to Dasharo Release Notes for your hardware platform can be found in
...
+Dasharo Supporters receive firmware updates more frequently than the community.
+Number of updates per year depends on the number of Dasharo Subscriptions sold
+and the availability of other funding (e.g., NLNet, corporate sponsors,
+community donations) but is less than 2 updates per year. Dasharo Supporters
+updates are characterized by a changing patch version (`z`). Fixes and features
+introduced in Dasharo Supporters releases will also be available later in
+community releases. In short, being a Dasharo Supporter gives an early access to
+the newest features and fixes.
+
+## Community releases
```
macpijan, line 27: What about releases for customers, such as NovaCustom, or Protectli? Do we also add some explaination there? How do we define which platforms gets released in which release cadence?

Maybe this model makes sense, but are we ready to employ it? I do not see it happening yet, which means we want to code some rules which we do not use yet.

macpijan (commented):
I think it is a good start, but we need to handle all cases, and maybe use it in practice first?

docs/dev-proc/versioning.md
macpijan: Also, I'm wondering if dev-process is the best place for it. It sounds more like information for end-users, who should not really want to dive into the dev-process documentation to know this.

miczyg1: > What about releases for customers, such as NovaCustom, or Protectli? Do we also add some explaination there? How do we define which platforms gets released in which release cadence?

Right, it may need some clarifications that customer releases are not bound bu these rules.

> Maybe this model makes sense, but are we ready to employ it? I do not see it happening yet, which means we want to code some rules which we do not use yet.

Sort of. We want to get ready for it and hit sis one step. As you are pointing out your concerns the idea can be polished before introduced to real life.

> Also, I'm wondering if dev-process is the best place for it. It sounds more like information for end-users, who should not really want to dive into the dev-process documentation to know this.

Maybe not, but this is where the versioning was explained. But devs will also need to respect the versioning.

pietrushnic: > Right, it may need some clarifications that customer releases are not bound bu these rules.

Upcoming FAQ start to explain:
- Dasharo Comunity Support (MSI Z690 et.al)
- Dasharo Support Package (Protectli, Novacustom)
- Dasharo Enterprise Package

What we doing here we prepare for communicating our vision for those products. I think it is better that way since we can validate it with community. Otherwise concepts will stay in our heads not exposed and not validated.

> Sort of. We want to get ready for it and hit sis one step. As you are pointing out your concerns the idea can be polished before introduced to real life.

We want to announce it during DUG, so we have to be roughly prepared. If we will wait for being ready we will never become ready.

> Maybe not, but this is where the versioning was explained. But devs will also need to respect the versioning.

@macpijan where would you put explanation about versioning scheme? Or maybe above message should be split into scheme and scheme explanation in user documentation.

macpijan: > We want to announce it during DUG, so we have to be roughly prepared. If we will wait for being ready we will never become ready.

Then go ahead

> where would you put explanation about versioning scheme? Or maybe above message should be split into scheme and scheme explanation in user documentation.

Let it stay here right now

macpijan (approved):

PR #506, opened 2023-03-06
docs/osf-trivia-list/dasharo.md: add section about porting Dasharo

miczyg1:


pietrushnic (approved):
@miczyg1 let me know what do you think.

pietrushnic:
@miczyg1 make sense, do you want to write paragraph about that? Does it fit here? Maybe it is more Dasharo Community Support definition?

miczyg1:
> @miczyg1 make sense, do you want to write paragraph about that? Does it fit here? Maybe it is more Dasharo Community Support definition?

Could be a part of Dasharo Community Support definition, indeed. But this is subject to another PR I think? We can always extend the Dasharo porting section nonce we have a Dasharo Community Support definition.

pietrushnic:
@miczyg1 yes, "what is DCS" is in comment at this point. Let's finish what is already in progress and we can back to commented questions later.

pietrushnic (approved):

PR #511, opened 2023-03-08
docs/variants: remove double slashes

pietrushnic:


pietrushnic:
It looks like it works without problems.

PR #517, opened 2023-03-09
docs/unified-test-documentation/dasharo-security/20P-boot-menu.md: add file

Pokisiekk:


docs/unified-test-documentation/dasharo-security/20P-boot-menu.md
```diff
@@ -0,0 +1,97 @@
+# Dasharo Security: Boot menu enable/disable
```
sulewskiprzemyslaw, line 1: Is the test suite nomenclature corresponds to the accepted product ID nomenclature?

docs/unified-test-documentation/dasharo-security/20P-boot-menu.md
```diff
@@ -0,0 +1,97 @@
...
+
+1. While the DUT is booting, the prompt
+   `BOOT_MENU_KEY to enter Boot Manager Menu` should be displayed.
+   Example of unwanted output:
+
+    ```bash
+    F11 to enter Boot Manager Menu
+    ```
+
+1. After using the `BOOT MENU KEY` during boot, the boot menu should not be
```
sulewskiprzemyslaw, line 96: @Pokisiekk And what happens in such a situation? Will we be stuck at the SOL level?

sulewskiprzemyslaw (changes requested):

docs/unified-test-documentation/dasharo-security/20P-boot-menu.md
sulewskiprzemyslaw, line 96: Maybe we should describe this part of `expected result` section in other words? For example: `Despite using Boot menu key boot menu should not been shown and the device should be booted to the OS`.

sulewskiprzemyslaw (changes requested):

docs/unified-test-documentation/dasharo-security/20P-boot-menu.md
Pokisiekk: As for the Product-id list, I didn't find such a module there, so I named it as analogous to other modules as possible.

docs/unified-test-documentation/dasharo-security/20P-boot-menu.md
Pokisiekk: I changed this point a bit. Mention of booting to the OS I would let go because the system isn't always installed.

sulewskiprzemyslaw (approved):

PR #528, opened 2023-03-27
Add SOL string information

miczyg1:


pietrushnic:
@macpijan IIUC, it would be good if @wessel-novacustom would take a look at this UI,

wessel-novacustom:
I already did and confirmed via Jira.

PR #567, opened 2023-05-19
docs/variants/protectli_vp2420/building-manual.md: update process

Pokisiekk:


macpijan:
This MR can be replaced with: https://github.com/Dasharo/docs/pull/570

The proposed changes do not fix the building procedure. We should no longer be using zip package since we have the blobs repository.

PR #573, opened 2023-05-26
docs/dasharo-tools-suite/documentation.md: add local fw update subsection & notes about warnings

zollerd:
Thank you @miczyg1 in `#dasharo-support:matrix.org` for clarification on the `flashrom` warning/notification.

I am not certain if the flashrom warning

```text
Setting BIOS Control at 0xdc from 0x8b to 0x89 failed.
New vaule is 0x8b.
```

is only applicable to my device, which is NV40 TGL, or if it is more general. In the former case that should perhaps be clarified in the docs as well.

Furthermore, I'm not sure how a local update of the EC firmware would be done (in line with the a local BIOS update as described here). In case I happen to investigate this in the future, I will add it.

### DIFF
--- a/docs/dasharo-tools-suite/documentation.md
+++ b/docs/dasharo-tools-suite/documentation.md
@@ -256,6 +256,7 @@ This section describes the functionality of the Dasharo Tools Suite. These are:
 * [Dasharo zero-touch initial deployment](#dasharo-zero-touch-initial-deployment),
 * [HCL Report](#hcl-report),
 * [Firmware update](#firmware-update),
+    - [Local firmware update](#local-firmware-update),
 * [EC transition](#ec-transition),
 * [EC update](#ec-update),
 * [additional features](#additional-features),
@@ -380,6 +381,14 @@ used for flashing. You can compare them with the values listed in the supported
 hardware section on docs.dasharo.com. Both these questions can be accepted by
 typing `Y`.
 
+If you see the following warning during the process, you do not need to worry
+about it:
+
+```text
+Warning: Setting BIOS Control at 0xdc from 0x8b to 0x89 failed.
+New vaule is 0x8b.
+```
+
 Procedure execution ends automatically on the reboot of the platform (unless it
 requires otherwise). After restarting the device, you can enjoy the updated
 version of Dasharo, which we provide for given hardware.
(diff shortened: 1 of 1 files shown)


zollerd:
Oh wow, `codespell` actually looks at the content of the fenced block of type `text`, and thinks the string `vaule` it finds in there should read `value`...

docs/dasharo-tools-suite/documentation.md
```diff
@@ -380,6 +381,14 @@ used for flashing. You can compare them with the values listed in the supported
 hardware section on docs.dasharo.com. Both these questions can be accepted by
 typing `Y`.
 
+If you see the following warning during the process, you do not need to worry
+about it:
+
+```text
+Warning: Setting BIOS Control at 0xdc from 0x8b to 0x89 failed.
+New vaule is 0x8b.
```
miczyg1, line 389:
```suggestion
New value is 0x8b.
```

docs/dasharo-tools-suite/documentation.md
```diff
@@ -443,6 +452,33 @@ Rebooting in 5s:
...
+
+```bash
+sudo flashrom -p internal --ifd -i bios -w [path/to/your/coreboot.rom]
+```
+
+If `flashrom` outputs the following, you do not need to worry about it:
+
+```text
+Enabling flash write... Warning: Setting BIOS Control at 0xdc from 0x8b to 0x89 failed.
+New vaule is 0x8b.
```
miczyg1, line 478:
```suggestion
New value is 0x8b.
```

miczyg1:
@zollerd looks good to me. All we need now is the commits to be signed, otherwise, the repository rules do not allow merging.
https://docs.github.com/en/authentication/managing-commit-signature-verification/adding-a-gpg-key-to-your-github-account

miczyg1:
> > All we need now is the commits to be signed
> 
> Sure thing.
> 
> > ```
> > - New vaule is 0x8b.
> > + New value is 0x8b.
> > ```
> 
> Since that was a literal quote from `flashrom`'s output, I'd have thought fixing `codespell` rules to not peek into `text` type fenced block would be preferable, but doesn't really matter I guess. However, since you added a signed commit, I can't rewrite the history with my signed commits without destroying your signature. I'll squash your commit into the previous one, I hope that's ok.

I think it has been fixed in some flashrom revision, because when I tried on my machine, the typo was not present.

PR #639, opened 2023-08-21
MSI recovery.md WSON8 probe specify 8x6mm

recallmenot:
WSON8 exists in 8x6 and 6x5mm packages.  
Part number on the image of the BIOS flash IC  is 25q256jweq, as per [datasheet](https://eu.mouser.com/datasheet/2/949/w25q256jw_spi_revd_09042018-1489579.pdf) page 93 that is an 8-pad WSON 8x6mm.

Or are there MSI boards that use the smaller 6x5mm package?


miczyg1:
> Or are there MSI boards that use the smaller 6x5mm package?

@recallmenot It should be WSON8 8x6mm only. Chip can either be Macronix or Winbond

miczyg1 (approved):

PR #644, opened 2023-09-01
MSI / NVC FW update mode

mkopec:
Use the (experimental) Annotations feature introduced in mkdocs-material v9.2.0


mkopec:
This is what it looks like: 
image
The idea is to avoid adding unnecessary verbosity and keep the document concise.

docs/unified/msi/firmware-update.md
```diff
@@ -14,22 +14,39 @@ Alternatively, it can be checked in the `BIOS Setup Menu`.
...
+
+1. v0.9.0 for Z790-P and v1.1.2 for Z690-A introduced support for Firmware
+   Update Mode
+
+=== "Firmware Update Mode"
+    Recent Dasharo releases come with Firmware Update Mode, an option in the
+    setup menu that disables all firmware protections for the duration of one
+    boot, during which firmware can be updated. To enter Firmware Update Mode,
+    open the [Dasharo Security Options](../../dasharo-menu-docs/dasharo-system-features.md)
+    menu and select the `Firmware Update Mode` option.
```
miczyg1, line 29: This should be rather moved to a generic place where we could link to from any board that support this option.

I would avoid doing the tabs for such purpose

requirements.txt
```diff
@@ -1,5 +1,5 @@
-mkdocs==1.4.1
-mkdocs-material==8.5.7
+mkdocs==1.5.2
+mkdocs-material==9.2.6
```
miczyg1, line 2: Such changes should rather go into separate PR and undergo more thorough verification if something broke.

docs/unified/msi/firmware-update.md
```diff
@@ -14,22 +14,39 @@ Alternatively, it can be checked in the `BIOS Setup Menu`.
...
 Please use one of the following environments to perform Dasharo update:
 
 * [Dasharo Tools Suite (DTS)](#dasharo-tools-suite)
 * [Linux distribution of your choice](#linux-distribution-of-your-choice)
 
-We recommend using the DTS.
+We recommend using DTS for updating firmware.
 
-### Dasharo Tools Suite
+#### Dasharo Tools Suite
```
miczyg1, line 49: Please do not change the header levels if possible. There may be some links depending on it elsewhere and I don't know if it will work. At least only the highest levels are linkable in mkdocs TOC.

miczyg1 (changes requested):

requirements.txt
mkopec: Alright, moved that change to https://github.com/Dasharo/docs/pull/645 and changed this PR's base to that PR.

docs/unified/msi/firmware-update.md
macpijan: @miczyg1 Do you have any specific location in mind?

mkopec:
Added a diagram using the mermaid.js plugin for mkdocs.

docs/unified/msi/firmware-update.md
miczyg1: We can go either with `docs/guides/firmware-update.md` or `docs/osf-trivia-list/firmware-update.md` (this is where generic flashrom documentation went), where we could consolidate some generic procedure how to perform firmware update, including enabling FUM.

mkopec: Moved to `docs/guides/firmware-update.md` as suggested in commit https://github.com/Dasharo/docs/pull/644/commits/d7fe7f67e6c923c1f237ef2867d59a996aecf74b

docs/guides/firmware-update.md
```diff
@@ -0,0 +1,98 @@
...
+    subgraph bootmanager [PlatformBootManagerLib]
+    bm1{{Firmware Update Mode EFI variable found?}}
+    bm1 -- False --> bm2[Continue booting]
+    bm1 -- True --> bm3[Remove Firmware Update Mode EFI variable]
+    bm3 --> bm4[Display Firmware Update Mode warning dialog]
+    bm4 --> bm5[Confirm user presence]
+    bm5 --> bm2
+    end
+
+    bootmanager --> userpresence{{User presence confirmed?}}
```
miczyg1, line 95: if FUM variable is not found then we should indeed continue booting. However the way it is shown on diagram it looks like user presence confirmation is checked on both paths on FUM and non-FUM.

The way I would see it:

If FUM variable is not set or not found, then go to "Continue booting" -> kill (no more arrows)

If FUM variable is set:

Remove Firmware Update Mode EFI variable -> 
Display Firmware Update Mode warning dialog ->
Confirm user presence ->
User presence confirmed?
-> (True) Boot OS
-> (False) Reboot (maybe an arrow should go back to the first reboot right before coreboot)
kill (no more arrows)

I would not add anything more outside PlatformBootManagerLib block.

docs/guides/firmware-update.md
```diff
@@ -0,0 +1,98 @@
...
+    setup4[Choose Firmware Update Mode]
+    setup5[Press ENTER to continue]
+    setup1 --> setup2 --> setup3 --> setup4 --> setup5
+    end
+    setup --> reboot[Reboot]
+    reboot --> coreboot
+    subgraph coreboot
+    cb1{{Firmware Update Mode EFI variable found?}}
+    cb1 -- False --> cb2[Enable firmware lockdown]
+    cb1 -- True --> cb3{{Firmware Update Mode EFI variable value?}}
```
miczyg1, line 69: That's a bit redundant to check if variable is found and then what is its value. I would simplify it to:

"Firmware Update Mode variable set to enabled?"

If true then follow FUM boot flow, otherwise (whether the variable was set to enabled or not found, doesn't;t matter, because every other case will follow the False branch)

docs/guides/firmware-update.md
```diff
@@ -0,0 +1,98 @@
...
+    cb3 -- False --> cb2
+    cb3 -- True --> cb4[Skip firmware lockdown]
+    cb2 --> cb5{{Boot payload}}
+    cb4 --> cb5
+    end
+
+    coreboot --> securitypkg
+    subgraph securitypkg [SecurityPkg]
+    sec1{{Firmware Update Mode EFI variable found?}}
```
miczyg1, line 78: That is not right. The FUM variable presence is not enough to alter the settings. We should check for the value. Ideally if you could apply my previous comment.

docs/guides/firmware-update.md
```diff
@@ -0,0 +1,98 @@
...
+    sec1 -- False --> sec2[Enable Secure Boot]
+    sec1 -- True --> sec3[Skip enabling Secure Boot]
+    sec2 --> sec4[Continue booting]
+    sec3 --> sec4
+    end
+
+    securitypkg --> bootmanager
+    subgraph bootmanager [PlatformBootManagerLib]
+    bm1{{Firmware Update Mode EFI variable found?}}
```
miczyg1, line 87: That is not right. The FUM variable presence is not enough to alter the settings. We should check for the value. Ideally if you could apply my previous comment.

docs/guides/firmware-update.md
```diff
@@ -0,0 +1,98 @@
...
+- SPI flash write-protection - prevents overwriting the initial bootblock and
+  Verified Boot code
+- SMM BIOS write protection - prevents all writes to BIOS flash memory outside
+  of privileged code running in System Management Mode
+- UEFI Secure Boot - in combination with Linux Kernel Lockdown, prevents direct
+  access to the SPI flash controller from the OS
+
+If you are interested, see the
+[Dasharo System Features](../dasharo-menu-docs/dasharo-system-features.md)
+article for more details.
```
miczyg1, line 25:
```suggestion
section for more details.
```

docs/guides/firmware-update.md
```diff
@@ -0,0 +1,98 @@
...
+![](./images/setup_menu_dsf.png)
+1. Navigate to `Dasharo System Features`:
+![](./images/setup_menu_dsc.png)
+1. Navigate to `Dasharo Security Options`:
+![](./images/setup_menu_lbm.png)
+1. Select the `Enter Firmware Update Mode` option:
+![](./images/setup_menu_fum.png)
+1. When prompted, press Enter to accept. The device will reboot in Firmware
+  Update Mode.
+1. When prompted, press the indicated key on the keyboard. Alternatively, to
```
miczyg1, line 44:
```suggestion
1. After reboot, when prompted, press the indicated key on the keyboard. Alternatively, to
```

miczyg1 (changes requested):

docs/guides/firmware-update.md
mkopec: That's the way it's implemented, at least for the Secure Boot portion in SecurityPkg https://github.com/Dasharo/edk2/commit/270f5b14260c6bfc266400427b4cdad3c7593157

docs/guides/firmware-update.md
mkopec: fixed in 957d3c3d2b4e2c1f9d38e70eb8f6fabd4463c196

docs/guides/firmware-update.md
mkopec: fixed in 957d3c3d2b4e2c1f9d38e70eb8f6fabd4463c196

docs/guides/firmware-update.md
mkopec: Updated in 957d3c3d2b4e2c1f9d38e70eb8f6fabd4463c196

docs/guides/firmware-update.md
mkopec: Updated in 957d3c3d2b4e2c1f9d38e70eb8f6fabd4463c196. I believe the mentioned logic in edk2 needs to be updated

docs/guides/firmware-update.md
mkopec: fixed in 957d3c3d2b4e2c1f9d38e70eb8f6fabd4463c196

docs/guides/firmware-update.md
mkopec: Updated in 957d3c3d2b4e2c1f9d38e70eb8f6fabd4463c196

> Reboot (maybe an arrow should go back to the first reboot right before coreboot)

Can't do that with mermaid.js, without having the layout of the flowchart changing entirely (platformbootmanagerlib moves to the top then)

docs/guides/firmware-update.md
miczyg1: Hmm. so it seems FUM variable just indicates the current state... I don't get it why on coreboot side we check the value as well. I missed it it seems. Well then, my bad, let's revert it to how it was before.

docs/guides/firmware-update.md
miczyg1: https://github.com/Dasharo/docs/pull/644#discussion_r1317212478

docs/guides/firmware-update.md
miczyg1: Here we do not need to revert it I think.

docs/guides/images/setup_menu_dsc.png
miczyg1, line 1: @mkopec please take all screenshots again in 80x25 mode. With such a high res, the menus are barely readable. You can do so by entering EFI Shell and issues a `mode` command with correct parameters than go back to setup by typing `exit` in the Shell.

docs/guides/firmware-update.md
miczyg1: @mkopec one arrow is still wrong. The true arrow from `Confirm user presence` should not start from `Confirm user presence` but from `User presence confirmed?`.  `Confirm user presence` is not a question, so no true/false arrow should start from there

docs/guides/images/setup_menu_dsc.png
mkopec: > You can do so by entering EFI Shell and issues a mode command with correct parameters than go back to setup by typing exit in the Shell.

Nice, thanks. Took the screenshots again in 80x25 mode.

docs/guides/firmware-update.md
mkopec: Switched the diagram to UML, please review again @miczyg1 @macpijan

docs/unified/msi/firmware-update.md
```diff
@@ -14,20 +14,35 @@ Alternatively, it can be checked in the `BIOS Setup Menu`.
...
+
+1. v0.9.0 for Z790-P and v1.1.2 for Z690-A introduced support for Firmware
+   Update Mode
+
+### Firmware Update Mode
+
+If the currently installed Dasharo version supports Firmware Update Mode, follow
+the steps outlined in
+[generic Firmware Update documentation](../../guides/firmware-update.md#firmware-update-mode).
+
```
BeataZdunczyk, line 29: @mkopec please add similar information to NovaCustom firmware update page: https://docs.dasharo.com/unified/novacustom/firmware-update/.

mkopec: Added

docs/guides/firmware-update.md
```diff
@@ -0,0 +1,53 @@
...
+1. Enter the Setup Menu:
+![](./images/setup_menu_dsf.png)
+1. Navigate to `Dasharo System Features`:
+![](./images/setup_menu_dsc.png)
+1. Navigate to `Dasharo Security Options`:
+![](./images/setup_menu_fum.png)
+1. Select the `Enter Firmware Update Mode` option:
+![](./images/setup_menu_fum_confirmation.png)
+1. When prompted, press Enter to accept. The device will reboot in Firmware
+  Update Mode.
```
miczyg1, line 43: The images seem to be off-by-one step. In point 2  we say to navigate to Dasharo SYstem Features, while we are already in Dasharo SYstem features on the picture below.

In point 4 we say to select Enter Firmware Update Mode, but in the picture below it we have already selected it and a prompt appeared. The information what to do about prompt is in next point...

I would merge point 1 and 2 and then shift all images one point lower

miczyg1 (changes requested):

docs/guides/firmware-update.md
miczyg1: I see we did a redesign of the diagram again... How am I supposed to close this thread?

Now I see that on the non-FUM path in the PlatformBootManagerLib we still have `User presence check failed?` which should not be there... There should be an arrow to `Boot` directly from `Do nothing`.

docs/guides/firmware-update.md
mkopec: Ok, done.

docs/guides/firmware-update.md
miczyg1: Also if we have Boot, why do we `Present boot menu` if FUm is not set? What boot menu? And why?

On non-FUM path it should be `Do nothing` -> `Boot` -> kill, that's it.

On FUM path we should have `Confirm User Presence` -> `User presence check failed?` -> (false) `Boot iPXE` -> `Boot DTS via iPXE` -> `Start Firmware Update`. Why is this so complicated?

mkopec: That's why I asked how much of the implementation details need to be here. Applied your suggestions in bfe41ce1931cbd57f2a1bb753bb84ae19e551328.

miczyg1: Now looks good. The arrows from `Set boto entry to iPXE` to `Boot` look ugly, but I guess we can't do much about it, it's just plain UML....

miczyg1 (approved):

PR #667, opened 2023-09-29
Add new MSI options

miczyg1:


mkopec:
Sorry if this was explained elsewhere, but what happens when the RAM sticks don't have XMP profiles or only have XMP1 and the user selects XMP2/3? Has this been tested? @miczyg1 @SergiiDmytruk

miczyg1:
> Sorry if this was explained elsewhere, but what happens when the RAM sticks don't have XMP profiles or only have XMP1 and the user selects XMP2/3? Has this been tested? @miczyg1 @SergiiDmytruk

You would have to check what FSP does in such case @mkopec

SergiiDmytruk:
> Sorry if this was explained elsewhere, but what happens when the RAM sticks don't have XMP profiles or only have XMP1 and the user selects XMP2/3? Has this been tested?

When I tried XMP2 on Z790-P it just didn't POST (I think I waited long enough).  I also saw `coreboot` print some error code from FSP once, it was during an automated test which probably went wrong, so not sure if it was XMP3 (also on Z790-P).

SergiiDmytruk (approved):

PR #678, opened 2023-10-13
variants/qemu_q35/releases.md: v0.1.0 scope

macpijan:


docs/variants/qemu_q35/releases.md
```diff
@@ -3,15 +3,39 @@
...
+- Serial Port Configuration menu
+- [Wi-Fi / Bluetooth module disable option in setup menu](https://docs.dasharo.com/dasharo-menu-docs/dasharo-system-features/#dasharo-security-options)
+- [Built-in webcam disable option in setup menu](https://docs.dasharo.com/unified/novacustom/camera-switch/)
+- [Battery threshold options in setup menu](https://docs.dasharo.com/dasharo-menu-docs/dasharo-system-features/#power-management-options)
+- [PCIe Configuration menu](https://docs.dasharo.com/dasharo-menu-docs/dasharo-system-features/#pcipcie-configuration)
+- [Memory configuration menu](https://docs.dasharo.com/dasharo-menu-docs/dasharo-system-features/#memory-configuration)
+
+### Binaries
+
+https://github.com/Dasharo/edk2/releases
```
macpijan, line 37: The idea would be to publish releases simply via github release for lowest effort. Let me know what you think of that.

docs/variants/qemu_q35/releases.md
```diff
@@ -3,15 +3,39 @@
...
 
-- Enabled TPM Support.
-- Secure Boot Support for Q35 machine.
-- Option to reset Secure Boot keys to defaults.
-- Dasharo System Features.
-- Dasharo setup password.
-- OPAL disk password support.
-- SATA disk password support.
-- SMM Feature enabled.
+- Configurable boot order
```
macpijan, line 10: I have gathered the features we had + enabled the ones not enabled in: https://github.com/Dasharo/edk2/pull/92

The thing is, many of them are just menus with no proper backed, as far as I understand.

Moreover, we will not test them properly right now (even if they would work), as in this v0.1.0 the main goal would be to have just menus and being able to test the menus (for correct strings, default values, moving around in the menus).

What can be "properly" tested right now is:
- custom boot menu keys
- UEFI shell

At least I have confirmed those 2 tests from our suite can pass.

docs/variants/qemu_q35/releases.md
pietrushnic: @macpijan I'm good with that, although I want to ensure we have the same quality criteria as regular Dasharo releases (public test results, hashes, signatures, release notes), but on GH.

docs/variants/qemu_q35/releases.md
pietrushnic: @macpijan that is fine if we clearly say that in the release notes. I guess one of the big things should be UEFI Secure Boot and DTS validation. For the further one, I would expect to create the correct backend on the QEMU side to handle various cases correctly.

Creating the correct backend for other features does not make sense for us now, but there is huge potential in confidential computing, which could fuel efforts to improve what we do for QEMU Q35, it probably would mean a slightly different goal and maybe QEMU target but fundamental operating principles would be the same.

docs/variants/qemu_q35/releases.md
macpijan: For signatures we would need to add signing key to github, which might not be the best idea.
Or maybe there is a way to add signatures to release later.

pietrushnic: @macpijan it depends on what signature you talking about - we had this discussion with @miczyg1 elsewhere. What I have in mind is a hash signing what is standard for us now in all releases and happen out of any CI.

macpijan: Yes, hash signing. If this should be part of CI, key must be stored somewhwere as well.

pietrushnic: @macpijan did I suggest that signing should be part of CI? None of our projects use automatic signing in CI, so why we would do it here? The signing could be a manual and the last step in the whole release process. When signing, the key owner would look at test results to confirm that the given hash contains given properties. Part of the CI can be binary publishing, signing should happen separately.

macpijan: Let's try that.

docs/variants/qemu_q35/releases.md
macpijan: @pietrushnic see my proposal of dividing these in: https://github.com/Dasharo/docs/pull/678/commits/45e294d357c78168eac9c27ddd09818f09dc4794

pietrushnic: @macpijan I love it. Your proposal is practical, and we clearly see where we are. What is more, we can build further roadmap on this information. IMHO it it good enough to close this discussion.

miczyg1:
LGTM overall, of course TBDs have to be replaced with correct values

pietrushnic (approved):

PR #677, opened 2023-10-10
dev-proc: update for new branch structure

mkopec:

PR #684, opened 2023-10-22
Qemu q35 v0.1.0 rc1

pietrushnic:
Fixes https://github.com/Dasharo/dasharo-issues/issues/541


docs/variants/qemu_q35/building-manual.md
```diff
@@ -1,66 +1,88 @@
...
+git clone https://github.com/Dasharo/edk2.git -b REVISION
 ```
 
-Clone official Dasharo EDK2 repository to your docker instance,
-with git or downloading the source code from github.
+Clone Dasharo EDK II Platforms fork:
 
 ```bash
-git clone https://github.com/Dasharo/edk2.git
+git clone https://github.com/Dasharo/edk2-platforms.git -b REVISION
```
miczyg1, line 40: We do not develop edk2-platforms at all basically, so we don't plan to tag it

miczyg1: If you wish to, you may tag this commit https://github.com/Dasharo/edk2-platforms/commit/3323ed481d35096fb6a7eae7b49f35eff00f86cf
with some semver for example.

pietrushnic: @miczyg1 the point is we have to tag because otherwise, we don't have a stable reference point, and we should have stable reference points for all firmware releases and all components. This is required for SBOMs and validation purposes.

> If you wish to, you may tag this commit https://github.com/Dasharo/edk2-platforms/commit/3323ed481d35096fb6a7eae7b49f35eff00f86cf
with some semver for example.

If you look at the commit, you can see it is tagged with `qemu_q35_v0.1.0-rc1`. I would like to move with what we have now because we are waiting to finalize #677 may block release progress for some time.

miczyg1: I would drop the platform in the tag name, because edk2-platforms usage in Dasharo is not related to given platform.

pietrushnic: @miczyg1 it is not related right now. Creating multiple tag schemes adds a maintenance burden and unnecessarily complicates procedures for end users. What should be a tagging scheme for edk2-platforms?

My goal is to achieve consistency in documentation from developers' and users' perspectives. If you think this is not a valid claim and we should apply different tagging schemes or not tag some repositories at all, we should fix that as part of #677 and treat the solution proposed here as temporary.

miczyg1: > What should be a tagging scheme for edk2-platforms?

Like I said, we do not actively develop edk2-platforms, so putting a tag like v1.0 and leaving it alone is sufficient

pietrushnic: IMHO, this unnecessarily breaks the scheme. But I fixed here is the fix: c2c9b94e750b

P.S. I don't like v1.0; that's why I added v0.9.0

docs/variants/qemu_q35/development.md
```diff
@@ -125,3 +125,112 @@ System Features. The below image shows the Dasharo System Features enabled for
...
+
+#### Build
+
+Start the instance of the docker image under the Dasharo/edk2 repository:
+
+```bash
+docker run --rm -it -v $PWD/../edk2:/home/coreboot/edk2 \
+    -v $PWD/../edk2-platforms:/home/coreboot/edk2-platforms \
+    -v $PWD:/home/coreboot/ipxe \
+    -w /home/coreboot/ipxe coreboot/coreboot-sdk:2022-12-18_3b32af950d \
```
miczyg1, line 177:
```suggestion
    -w /home/coreboot/ipxe coreboot/coreboot-sdk:2021-09-23_b0d87f753c \
```

I guess?

docs/variants/qemu_q35/development.md
```diff
@@ -125,3 +125,112 @@ System Features. The below image shows the Dasharo System Features enabled for
...
+$EDK2_PLATFORMS_PATH/Features/Intel/OutOfBandManagement:\
+$EDK2_PLATFORMS_PATH/Features/Intel/PowerManagement:\
+$EDK2_PLATFORMS_PATH/Features/Intel/SystemInformation:\
+$EDK2_PLATFORMS_PATH/Features/Intel/UserInterface"
+```
+
+To build the image simply invoke the following command
+
+```bash
+build -a IA32 -a X64 -t GCC5 -b RELEASE -p OvmfPkg/OvmfPkgX64.dsc
```
miczyg1, line 235: So we have two separate build instructions:
1. Building manual
2. Development

docs/variants/qemu_q35/initial-deployment.md
```diff
@@ -13,3 +13,11 @@ qemu-system-x86_64 -machine q35,smm=on \
 	-debugcon file:debug.log -global isa-debugcon.iobase=0x402 \
 	-global ICH9-LPC.disable_s3=1
 ```
+
+If you use system without graphical output you may faces following issues:
```
miczyg1, line 17:
```suggestion
If you use system without graphical output you may face following issues:
```

docs/variants/qemu_q35/development.md
pietrushnic: @miczyg1 I forgo to push 7aaa4cd6be78

docs/variants/qemu_q35/development.md
pietrushnic: @miczyg1 those are almost the same, but Building Manual is not modular enough to scale into this. TBH, I can remove the whole section from here; those were mostly development notes while enabling networking. If that causes trouble in the review, I would rather remove the whole section than waste our time.

miczyg1: My concern is that we should rather unify things, not divide and duplicate/repeat (which you also dislike IIRC).

pietrushnic: @miczyg1 in that case, it would be redesigning the building manual for all supported platforms, which is out of scope here. Knowing how much can be improved, we can easily achieve infinite scope creep - it looks like the removal of old AMD from the top of the coreboot tree. I pointed to inconsistencies in the building manual two years ago: https://github.com/Dasharo/dasharo-issues/issues/37

We had no issues merging tons of documentation even though the problem was present. I wonder why we apply different standards right now.

To satisfy your comment, I will remove that section from the document. I'm concerned about the development section at all, but it was approved at some point, so I guess it meets your or someone else quality criteria. Fix:

Let's also think about how we can accelerate time to market - meaning release delivery instead of delaying it forever by repeatedly nitpicking. I recommend reading [this](https://blog.danlew.net/2021/02/23/stop-nitpicking-in-code-reviews/) because perfect is often the enemy of good.

miczyg1: I guess we still don't understand each other. My point its:

Why do we need separate build instructions for development and build manual?
 
If we want to achieve the same functionality iPXE should be always built, not matter dev or not, so... Includig iPXE build procedure in build manual should be the right thing to do. Build manual always describes the procedure to build the same binaries as the release binaries. The same should apply here. 

> We had no issues merging tons of documentation even though the problem was present. I wonder why we apply different standards right now.
 
No idea what you mean. I simply want to follow the same standards and say that you don't need separate build instructions...

pietrushnic: > I guess we still don't understand each other here @pietrushnic . My point its:

@miczyg1 definitely.

> Why do we need separate build instructions for development and build manual?

We don't. I removed separate building instructions in the commit I pointed to https://github.com/Dasharo/docs/commit/8a5e23cfbd71728179605db9d735c229852e2ed0

> If we want to achieve the same functionality iPXE should be always built, not matter dev or not, so... Includig iPXE build procedure in build manual should be the right thing to do. Build manual always describes the procedure to build the same binaries as the release binaries. The same should apply here.

This is precisely the state of the commitment right now. Please point to the precise line which should be changed, because you discuss something that does not exist now.

> No idea what you mean. I simply want to follow the same standards and say that you don't need separate build instructions...

Previously you wrote:

> My concern is that we should rather unify things, not divide and duplicate/repeat (which you also dislike IIRC).

So please take a look at the building manuals. Those are not unified. If you want to avoid dividing, duplicating/repeating, please do not merge building manual instructions that are not unified. IMO, in this discussion, different rules are applied because previously not, unified documentation was merged, but now when we had some duplication (one line, what was already fixed), we apply some different rules than the unification rule you explain. I just point out inconsistencies in the review process. I will try to address those in a better suited place than this thread. Let's focus on the code.

miczyg1: > So please take a look at the building manuals. Those are not unified. 

Yes they aren't. But we can slowly work towards that in separate efforts elsewhere. ANyways, my point was addressed to not duplicate building instructions in the same platform boundary. Closing.

miczyg1 (approved):

PR #685, opened 2023-10-25
New secure boot

miczyg1:


docs/images/menus/secure_boot_main3.png
macpijan, line 1: @miczyg1 So by default the `Enable Secure Boot` is still grayed out option and I need to enroll keys first?

From the perspective of test automation, this layout does not look simpler to handle, but rather harder to work around. So on the serial I would have `Enable Secure Boot` which can be inactive, but still present on serial as a menu entry. And I will also have 2 additional strings in the options are, which are not really selectable options.

miczyg1: > So by default the Enable Secure Boot is still grayed out option and I need to enroll keys first?

No.

```
By default, Dasharo firmware boots with the default keys/certificates enrolled
(see [Custom mode and key management](#custom-mode-and-key-management)) and
default UEFI Secure Boot state depending on platform (in most cases disabled).
```

You will see such image only if your erase the SB keys or miss PK. @macpijan

miczyg1: Does this sentence help a bit? https://github.com/Dasharo/docs/pull/685/files#diff-a4b2d33002947c75ecf3ffa3a32682f2d33b94d57fb13a700b1f8acecb625dfeR91

miczyg1: @pietrushnic https://github.com/Dasharo/docs/pull/685/files#diff-a4b2d33002947c75ecf3ffa3a32682f2d33b94d57fb13a700b1f8acecb625dfeR108

pietrushnic: @miczyg1 :heart: but we should plan the inclusion of Qubes OS Certificates, although I'm not sure when they plan to implement UEFI Secure Boot.

miczyg1: We can think about it when there is some timeline from them

macpijan: > Only if you mess with the keys in advanced menu.

But we will have to test that flow. And we can make our lives testing easier or harder, keeping the UX clean, IMHO.

macpijan: I can see you have added `>` to selectable entries, which is great and will make things easier.
The only remaining comment would be why the `Current Secure Boot State` is selectable entry.

I think this is literally the only single one in our menus, which can be selected, but there is no action you can do with that.

miczyg1: `Current Secure Boot State` is not selectable. It is just informational label about the state and has never been touched since the beginning of our edk2 journey @macpijan

miczyg1: @macpijan anything else needed to close this one?

macpijan: Nothing right now. Created issue for fruther consideration: https://github.com/Dasharo/dasharo-issues/issues/549

miczyg1:
@macpijan @pietrushnic anything to change in the PR, or can we proceed with merging?

pietrushnic:
@miczyg1 I don't have a target to test those changes. Unfortunately, I cannot confirm those changes.

miczyg1:
> @miczyg1 I don't have a target to test those changes. Unfortunately, I cannot confirm those changes.

QEMU?



> @macpijan @pietrushnic anything to change in the PR, or can we proceed with merging?

@macpijan ?

macpijan (approved):

PR #693, opened 2023-10-29
Add another HCL entry for MSI

xbjfk:
Can confirm that with my GPU I can access the setup menu. PCI id is `1002:731f` - reference model 5700 XT.


pietrushnic:
@xbjfk thank you for your contribution.

PR #695, opened 2023-10-30
variants/novacustom: add missing dev_signed binaries

macpijan:


mkopec:
@macpijan I think I missed this PR when I submitted https://github.com/Dasharo/docs/pull/696

Do we need to publish these binaries to end users? I guess their biggest value for users is their reproducibility, but the links can be hidden somewhere less visible IMO. We've had an user ask which type of binaries they should download for their NV4x ADL.

PR #696, opened 2023-10-30
variants/novacustom_nv4x_adl/releases.md: remove link to dev_signed bin

mkopec:
dev_signed binaries are used for testing and development, but are not relevant for regular users and introduce unnecessary confusion. A user has already asked which one they should use.

Remove the link for clarity.


miczyg1:
This is not consistent with what we do for MSI. dev_signed are needed to confirm reproducibility, so at least we should level the signatures and hashes.

mkopec:
Ack. Would be nice to have a tool to compare code sections of different binaries, ignoring data sections like ME data, bootsplash, Vboot sigs, MRC cache etc

macpijan:
> and introduce unnecessary confusion. A user has already asked which one they should use.

The need of explaining this to users is still unresolved I assume.

> dev_signed are needed to confirm reproducibility

How do we communicate that to  users?

PR #719, opened 2023-12-09
Update firmware-update.md

wessel-novacustom:
Explained the order of firmware flashing and noted the requirement of network boot enabled.


macpijan:
Thanks for your contribution @wessel-novacustom !

I have picked that change in: https://github.com/Dasharo/docs/compare/add-info-on-enabling-network-boot?expand=1 fixing the pre-commit errors.

PR #732, opened 2024-01-02
dasharo-menu-docs/device-manager.md: update to include MS 2023 SB certs

mkopec:
add latest certs from https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/windows-secure-boot-key-creation-and-management-guidance


mkopec:
Now I'm wondering if we're actually including the new MS 2023 KEK, I only know for sure we updated DB

miczyg1 (approved):

miczyg1:
> Now I'm wondering if we're actually including the new MS 2023 KEK, I only know for sure we updated DB

KEK is there as well: https://github.com/Dasharo/edk2/blob/dasharo/UefiPayloadPkg/UefiPayloadPkg.fdf#L307

PR #746, opened 2024-01-22
docs/variants/protectli_vp66xx: Add initial documentation

miczyg1:


macpijan (approved):
Thanks. Good enough for now, but we should look into how we can have more common documentation between all/most Protectli boards. The recovery procedure is in progress already: https://github.com/Dasharo/docs/pull/753

PR #756, opened 2024-02-02
docs/dasharo-tools-suite/releases.md: hide earlier releases

WiktorG351:
This PR does two things:
* comments out all the earlier releases, so that they aren't visible on the page
* makes the `Latest release` section bold, so that users have no choice but to go to the releases page and download the latest version


TomaszAIR (approved):

pietrushnic:
@TomaszAIR, we could do better than what was implemented here. IMHO w should remove old releases, there is no value in those, except DTS archeology. Nobody should use those, and nobody will be supported when using those. So either we archive the section with old releases or remove it.

PR #782, opened 2024-03-11
docs/variants/protectli_vp46xx: document ath10k firmware issue

pkubaj:
Signed-off-by: Piotr Kubaj <piotr.kubaj@3mdeb.com>

### DIFF
--- a/docs/variants/protectli_vp46xx/overview.md
+++ b/docs/variants/protectli_vp46xx/overview.md
@@ -37,4 +37,4 @@ For more information please refer to the references below.
     validation procedure.
 * [Test matrix](test-matrix.md) - describes validation scope used during
     Dasharo firmware validation procedure.
-* [Post-installation setup](post-install.md) -
+* [Post-installation setup](post-install.md)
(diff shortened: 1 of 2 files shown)


docs/variants/protectli_vp46xx/overview.md
```diff
@@ -37,3 +37,4 @@ For more information please refer to the references below.
     validation procedure.
 * [Test matrix](test-matrix.md) - describes validation scope used during
     Dasharo firmware validation procedure.
+* [Post-installation setup](post-install.md) -
```
macpijan, line 40: @pkubaj Either the `-` at the end is not required, or some description after should be added?

macpijan (changes requested):
I would fix this myself, it it would be a branch in dasharo repo, not a fork

docs/variants/protectli_vp46xx/overview.md
miczyg1:
```suggestion
* [Post-installation setup](post-install.md)
```

miczyg1:
> I would fix this myself, it it would be a branch in dasharo repo, not a fork

@macpijan there is a w/a to that. Simply use suggestion via GH UI and it lets you apply the suggestion on a fork :)

miczyg1 (approved):

PR #787, opened 2024-03-18
docs/guides/firmware-reflash.md: add flashing guide

mkopec:
Simple and generic. Of course sometimes the commands may be different depending on mainboard, version, branch etc, but this guide should cover most cases.


filipleple:
Looks good to me, covers all important matters

filipleple (approved):

PR #803, opened 2024-04-18
protectly_v*/building_manual.md: s/aof/of/

pkubaj:


pkubaj:
Fixed.

macpijan (approved):

pkubaj:
I'm also not sure whether "You should checkout to the same tag as in case of the coreboot repository" is a proper usage of "in case".

PR #817, opened 2024-05-16
Vp2420 re publish

mkopec:


BeataZdunczyk:
LGTM, but SEO CI is not happy yet. @mkopec

pietrushnic:
@BeataZdunczyk @mkopec I reported that internally, someone moved files without thinking that those are linked from docs.dasharo.com; it is good we have a mechanism to check that. The ideal would be to prevent such files from moving in the repo with binaries.

BeataZdunczyk (approved):
LGTM

PR #835, opened 2024-06-17
docs/dev-proc/standard-release-process.md: Add rel candidate step

mkopec:


docs/dev-proc/standard-release-process.md
```diff
@@ -13,15 +13,14 @@ should be maintained in that way.
    to understand versioning scheme please read [Versioning](versioning.md)
    section
 2. Merge current platform development branches to `<platform>_rel_vX.Y.Z`
-3. Run platform regression test suite
```
macpijan, line 16: We can simply use `1.` everywhere to avoid the need of renumbering whole lists.

macpijan (approved):

PR #838, opened 2024-06-19
Add overview of PCR measurements

SergiiDmytruk:


docs/kb/pcr-measurements.md
```diff
@@ -0,0 +1,157 @@
...
+measured in sequence.  Keep this in mind when looking for something in the event
+log as it can sometimes appear more than once, including several times in a row.
+
+#### Log entries changing values
+
+Bootloaders or components of operating systems might update SecureBoot
+configuration automatically thus changing PCR values.  Similarly, any kind of
+software update or even update of configuration can potentially change
+behavior and cause an unexpected PCR value.  In such cases PCR values tend to
+stability after a reboot.
```
miczyg1, line 149:
```suggestion
stabilize after a reboot.
```

miczyg1 (changes requested):

docs/kb/pcr-measurements.md
```diff
@@ -0,0 +1,157 @@
...
+[TPM12]: https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClientImplementation_1-21_1_00.pdf
+[TPM20]: https://trustedcomputinggroup.org/wp-content/uploads/TCG_PCClient_PFP_r1p05_v23_pub.pdf
+[coreboot-pcrs]: https://doc.coreboot.org/security/vboot/measured_boot.html#platform-configuration-registers
+
+## coreboot measurements
+
+### PCR banks
+
+At the time of writing (June 2024) coreboot has no support for multiple PCR
+banks but allows choosing a hash algorithm for TPM 2.0 log format.  This choice
```
krystian-hebel, line 49: I don't like the wording, multiple PCR banks (which don't really have anything to do with event log) can be enabled by coreboot, it just doesn't make use of them. One could argue that this poses a security issue, which EDK (kind of) deals with, see next comment.

docs/kb/pcr-measurements.md
```diff
@@ -0,0 +1,157 @@
...
+| TPM version | PCR Banks
+| ----------- | --------
+| 1.2         | SHA-1 only
+| 2.0         | SHA-256 only
+
+This has an implication for TPM 2.0 case when multiple PCR banks are active and
+used by EDK.  This is handled by using dummy hash values (`0100...`) for missing
+digests (SHA-1) which do not correspond to actual PCR updates.  Doing this
+breaks replaying of TPM log unless fake digests are skipped, but it allows to
+comply with TPM event log format.
```
krystian-hebel, line 61: This is done exactly to mark that given bank isn't used. On top of that, a dummy value (`EV_SEPARATOR`, which I believe is `0100...` mentioned in this paragraph) should be extended on PCRs from unused banks before execution is passed to user-controlled environment. Without it, an attacker could fake the PCR values by extending them from their defaults of all 0s, by replaying the event log on live PCRs.

If `EV_SEPARATOR` exists in the event log, but isn't extended in actual PCRs, it is a bug. See `8.2.4 Measuring OS Boot Events` in TPM2.0 spec linked earlier in the doc.

docs/kb/pcr-measurements.md
```diff
@@ -0,0 +1,157 @@
...
+
+### PCR measurements
+
+| PCR   | Event type | Condition  | Description
+| ---   | ---------- | ---------  | -----------
+| PCR-2 | EV_ACTION  | Always     | Stages, data, blobs, payload
```
krystian-hebel, line 67: It is worth mentioning that `data` refers to constant data, i.e. not configuration options which can change between boots on given FW version.

That said, are variables read by coreboot from SMMSTORE measured?

docs/kb/pcr-measurements.md
```diff
@@ -0,0 +1,157 @@
...
+| PCR-4    | EV_EFI_BOOT_SERVICES_APPLICATION | Boot menu/Setup | Hash of an EFI application implementing UI
+| PCR-1    | 0x00DA0000                       | First boot try  | Dasharo-specific EFI variables (NUL-terminated ASCII name followed by variable's value)
+| PCR-1    | EV_EFI_VARIABLE_BOOT             | First boot try  | 61dfe48b-ca93-d211-aa0d-00e098032b8c:BootOrder variable
+|          |                                  |                 | 61dfe48b-ca93-d211-aa0d-00e098032b8c:Boot0000... variables
+| PCR-4    | EV_EFI_ACTION                    | Any boot try    | "Calling EFI Application from Boot Option"
+| PCR-0..7 | EV_SEPARATOR                     | Always          | Separator
+| PCR-1    | EV_EFI_HANDOFF_TABLES            | First boot try  | SMBIOS base+size (could use v2 of the event, but doesn't ATM)
+| PCR-4    | EV_EFI_BOOT_SERVICES_APPLICATION | Any boot try    | Hash of an EFI application (bootloader, UEFI shell, etc.)
+| PCR-7    | EV_EFI_VARIABLE_AUTHORITY        | Unknown         | Measurements done by a bootloader (GRUB, shim, etc.), listed here to show where they appear
```
krystian-hebel, line 112: This is (should be?) part of SecureBoot, not the bootloader.

docs/kb/pcr-measurements.md
```diff
@@ -0,0 +1,157 @@
...
+
+[grub]: https://www.gnu.org/software/grub/manual/grub/html_node/Measured-Boot.html
+[shim]: https://github.com/rhboot/shim/blob/master/README.tpm
+
+## Caveats
+
+### Accuracy
+
+Exact list of measurements depends on the combination of firmware components,
+their versions, installed hardware, TPM version, runtime settings and at least
```
krystian-hebel, line 132: I think this is what you meant:

```suggestion
their versions, installed hardware, TPM version, runtime settings and at last
```

docs/kb/pcr-measurements.md
```diff
@@ -0,0 +1,157 @@
...
+behavior and cause an unexpected PCR value.  In such cases PCR values tend to
+stability after a reboot.
+
+#### Conditional measurements
+
+Some things are measured dynamically when they get used which in turn is
+controlled by some setting.  This can result in a single configuration change
+affecting event log in multiple places, for example: disabling a device will
+change value of a corresponding option and possibly remove an entry with a
+measurement of its OptionROM.
```
krystian-hebel, line 157: I would also briefly mention user actions (setup, boot menu) here. It seems to be one of the most common causes of unexpected values, at least in my experience.

krystian-hebel (changes requested):

docs/kb/pcr-measurements.md
SergiiDmytruk: https://github.com/Dasharo/docs/compare/d8644747bf6d813d99ac1d9f52c60a916ef47475..c0376d7b5c8c64f9124139a5bc44c20cdf472812#diff-c2a55b23fc3bd588ed98d40c36104af5b9175c3acdc2ab8c0288d1590992e689R48

docs/kb/pcr-measurements.md
krystian-hebel: Check description in Table 14, it has two separate cases for error and success for `EV_SEPARATOR`, `0100...` seems to be the error case, after endianess swap.

Convenience shouldn't take precedence over abiding to specification. It is better to make log replaying fail in this case, as it suggests security issue (non-capped PCRs that can be extended _after_ the chain was already broken, for given bank).

In any case, log by itself isn't a proof, you can't depend on its correctness unless root of trust isn't compromised **and** PCR values are known to be good, and not only matching the log. In other words, if PCRs match the event log, it means that event log is properly created, but it doesn't give any security guarantees. Here, we don't even match the "properly created" criteria.

docs/kb/pcr-measurements.md
krystian-hebel: > They aren't, but EDK measures them right before boot.

I'm surprised to see this, but it is allowed by the specification. Even security-relevant configuration data only `SHOULD` be measured, and it only has to be done before `EV_SEPARATOR` (on Ready to Boot event), not necessarily before the data is actually used. I can think of few attacks that may be able to make use of it.

docs/kb/pcr-measurements.md
SergiiDmytruk: > Check description in Table 14, it has two separate cases for error and success for EV_SEPARATOR, 0100... seems to be the error case, after endianess swap.

That is still about capping of PCRs, it's fixed to 4 bytes and I don't think endianness applies because measured data is a byte array.

> (non-capped PCRs that can be extended after the chain was already broken, for given bank).

It's capped by EDK which also adds fake digests to the log.

SergiiDmytruk:
A correction about first boot and when SecureBoot variables get measured: https://github.com/Dasharo/docs/compare/1761acc7b3e107fe860738d3a7536dd14e598f37..1022511ede55ff4dc935554f8c84f095ebffb3fc

docs/kb/pcr-measurements.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Caveats
+
+### First Boot
+
+First boot is somewhat special due to firmware initializing its state.  This
+includes SecureBoot-related variables which generate
+`EV_EFI_VARIABLE_DRIVER_CONFIG` events.  SecureBoot is considered to be on at
+the start, but than it's switched to its default off state.  As a result, event
+log on first and subsequence boots will differ even if no configuration is
```
m-iwanicki, line 135:
```suggestion
the start, but then it's switched to its default off state.  As a result, event
log on first and subsequent boots will differ even if no configuration is
```

SergiiDmytruk: Thanks. https://github.com/Dasharo/docs/compare/1022511ede55ff4dc935554f8c84f095ebffb3fc..e825eefa841d8614c5701d116dc8150a3432222a

krystian-hebel (approved):

docs/kb/pcr-measurements.md
krystian-hebel: @miczyg1 ping, your "changes requested" is blocking the merge.

miczyg1 (approved):

PR #865, opened 2024-07-25
Document battery behavior for novacustom

philipandag:


docs/unified/novacustom/features.md
```diff
@@ -50,12 +50,34 @@ battery state, the following error message will be displayed:
...
+
+The device can be powered in three modes:
+
+- Battery Mode - the device is powered using the battery only. This mode is
+active only if the device is not connected to a power adapter.
+- Charging Mode - when the device is connected to a proper power adapter
+it is powered directly from the adapter. The battery will start charging only
+when the current charge falls below the lower charge threshold. When the upper
+threshold is achieved the battery is physically disconnected from the system
+and does not experience any excessive wear. If you are using the device like a
```
mkopec, line 71: I wouldn't say it's physically disconnected, it's just not powering the system, or being charged. 

```suggestion
threshold is achieved, the battery stops charging and no current is flowing through it,
preventing excessive wear. If you are using the device like a
```

```suggestion
threshold is achieved the battery is physically disconnected from the system

```

philipandag: If this behavior is controlled using some high current transistor then that "physically" would not be true. Moreover it has to be connected to a battery management system, so I agree my claim was not accurate... Also, a battery is most certainly inanimate so it can't "experience" anything. Thanks for the suggestions!

mkopec (approved):

PR #866, opened 2024-07-29
kb/edk2-capsule-updates.md: add

SergiiDmytruk:
This provides some background information about capsules and their use in EDK2.

### DIFF
--- a/docs/kb/edk2-capsule-updates.md
+++ b/docs/kb/edk2-capsule-updates.md
@@ -1,0 +1,302 @@
+# Working with UEFI capsule updates in EDK2
+
+## Introduction
+
+The purpose of this document is to provide an overview of means for working with
+UEFI update capsules that can be employed by an EDK2-based firmware.  This is,
+however, not an exhaustive list of such tools and there are other ways of
+managing capsules as is alluded to by the last section.
+
+## Tools in EDK's source tree
+
+[`BaseTools/BinWrappers/PosixLike/GenerateCapsule`][GenerateCapsule] (really
+[`BaseTools/Source/Python/Capsule/GenerateCapsule.py`][GenerateCapsule.py]) can
+be used to generate/unpack/dump capsules.
+
+[`BaseTools/Source/Python/Pkcs7Sign`][Pkcs7Sign] can be used for signing
+without `GenerateCapsule` but test keys there are more interesting.  The keys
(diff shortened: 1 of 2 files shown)


BeataZdunczyk (approved):

docs/kb/edk2-capsule-updates.md
```diff
@@ -0,0 +1,302 @@
...
+
+`BaseTools/Source/Python/Pkcs7Sign` contains everything necessary for testing
+purposes and that's what will be used below.
+
+Real-world usage requires proper key management.  Key generation is covered by
+the README of `Pkcs7Sign` as well as in [tianocore's wiki][wiki-keygen].
+`Pkcs7Sign` additionally describes how to integrate generated keys into EDK2's
+build system.
+
+[wiki-keygen]: https://github.com/tianocore/tianocore.github.io/wiki/Capsule-Based-System-Firmware-Update-Generate-Keys
```
krystian-hebel, line 134: This mostly covers https://github.com/Dasharo/dasharo-issues/issues/806, we will only need to add user documentation there.

docs/kb/edk2-capsule-updates.md
```diff
@@ -0,0 +1,302 @@
...
+}
+```
+
+Some notes on JSON:
+
+- `GenerateCapsule` warns if `EmbeddedDrivers` isn't provided, having an empty
+  array if one doesn't need to embed any drivers is enough to silence the
+  warning
+- all (even numerical) fields must be given as strings or there will be errors
+- paths to files undergo expansion of environment variables
```
krystian-hebel, line 174: Only paths or can other fields be parametrized this way too? I'm mostly thinking about `FwVersion`, it would be tedious to change it with every release, OTOH we can `sed` this file on build time so it's not a big deal if it isn't possible.

SergiiDmytruk: Only paths to files or commands use `os.path.expandvars` to process input value: https://github.com/Dasharo/edk2/blob/eb218796794a6225f4d5813de29f49d865c7f9b1/BaseTools/Source/Python/Capsule/GenerateCapsule.py#L303-L316

If depending on `jq` is fine, it can be used as a more reliable alternative to `sed`:
```
jq '.Payloads[].FwVersion = "0x10000000"' dasharo.json
```

krystian-hebel (approved):

PR #869, opened 2024-07-30
Mv lab assembly to guides

pietrushnic:
One important thing for every platform that lands in the Dasharo Certification Lab is the Lab Assembly Guide. Maybe there should be a section Lab Assembly in Guides on the website, where all Lab Assembly Guides land, the current structure is a kind of complete Ecosystem->Transparent Validation-><Platform Name>-> Laboratory stand assembly guide. My proposal would be Guides->Lab Assembly-><Platform Name>


miczyg1 (approved):
Guides going to Guides section. +1

PR #820, opened 2024-05-21
[RFC] Initial documentation of Dasharo Product Naming Convention

pietrushnic:

PR #880, opened 2024-08-13
Dasharo Product Naming Convention - Pro Package change

Stanislaw-bnk:


pietrushnic:
Two things:
- it should probably depend on https://github.com/Dasharo/docs/pull/820/files
- https://github.com/3mdeb/3mdeb-website/pull/158#issuecomment-2285589545

Maybe we should use the Dasharo Pro/Enterprise Package? Or is this too convoluted?

artur-rs (approved):

artur-rs:
@Stanislaw-bnk such naming scheme change should be correctly communicated to the current and future users, maybe even the standalone S&M campaign for such changes

pietrushnic:
@artur-rs, it was communicated during DUG, so SM campaign or newsletter messaging could be prepared based on that.

Stanislaw-bnk:
@pietrushnic @artur-rs I would keep the Dasharo Pro Package without Enterprise, as that is the package we have listed in the store.

pietrushnic:
@Stanislaw-bnk I'm okay with that. We should explain it; if there are places where there is confusion, we could link to convention documentation. In conventional documentation, give a link to the updated TOS, which explains the difference in terms of support.

Stanislaw-bnk:
@artur-rs Seems like this is ready for merge? However, @pietrushnic your approval is still missing, as it’s currently `awaiting requested review from pietrushnic`.

As for the point `In conventional documentation, give a link to the updated TOS, which explains the difference in terms of support`, I see that there is no mention in the Dasharo Terms of Service differentiating between the Pro and Enterprise Packages, where PRO is only for personal use up to 5 units, and ENTERPRISE is for personal use above 5 units or for companies from the first unit. Shouldn't it be mentioned in the ToS? @gosiastyba

rafkoch:
@pietrushnic we need Your approval because now we see info: "requested review from pietrushnic"

pietrushnic:
@rafkoch That's okay, but please take a look at how many conflicts were created over that time. Those have to be resolved before I can approve.

pietrushnic:
Most of the improvements were merged in other places, so the difference after rebase is tiny. Thanks to it, I will merge it since other issues could be resolved before DUG#8.

PR #882, opened 2024-08-16
Improve ec recovery process documentation

philipandag:


macpijan:
@philipandag Do you think you can fix the linkchecker?

philipandag:
@macpijan I have updated the link and the list of features the new Dasharo Pro Package according to the subscription for Desktops from the shop. I think someone familiar with the our offer should verify this.

macpijan (approved):

PR #888, opened 2024-08-28
Add version matrix

arturkow2:


BeataZdunczyk:
@arturkow2 great, thank you.  QEMU is missing from the list.

docs/variants/versions.md
```diff
@@ -0,0 +1,29 @@
+# Version matrix
+
+## About
+
+The table below contains an up to date list of all supported platforms,
+including latest supported coreboot and EDK2 versions and branches from which
+you can build firmware for your platform.
+
+| Platform                | coreboot branch                                                                                       | Dasharo base                                                          | coreboot version | EDK2 version | EDK2 rev                                                                                    |
```
krystian-hebel, line 9: I think that `coreboot version` should be removed. It is strictly correlated with `Dasharo base`, and dropping this column will make more space for the rest.

krystian-hebel (changes requested):

docs/variants/versions.md
arturkow2: Done

krystian-hebel (approved):

krystian-hebel:
Merging despite CI complaining about broken link to https://review.coreboot.org/plugins/gitiles/memtest86plus/+/0bd34c22, caused by https://osfw.slack.com/archives/CCWLQKEHG/p1725964231625859

pietrushnic:
I saw the message on coreboot ml. Since there is no mirror, the only solution would be to remove the link if gitles will not be brought back. What would you say about mirroring that ourselves?

PR #907, opened 2024-09-25
Fix des keys to dpp

EduKav1813:
Addresses [issue 1046](https://github.com/Dasharo/dasharo-issues/issues/1046)


m-iwanicki (changes requested):
1. Why does your background in image looks kinda grey?
2. There is also DES mentioned in `Dasharo zero-touch initial deployment` section
    > option d stands for [DES](https://docs.dasharo.com/ways-you-can-help-us/#become-a-dasharo-entry-subscription-subscriber) release and it is only available to Dasharo Entry Subscription subscribers

    Current link is https://docs.dasharo.com/ways-you-can-help-us/#become-a-dasharo-pro-package-subscriber
3. There is still `https://docs.dasharo.com/osf-trivia-list/dts/#how-can-i-use-my-dasharo-entry-subscription-credentials` to fix (old naming + old interface gif)

EduKav1813:
> 1. Why does your background in image looks kinda grey?

I used an ssh connection to the qemu machine to get this screenshot. The terminal in the qemu itself did not have enough rows in it to capture all the output.

EduKav1813:
> Ad 2. Currently in markdown there is
> 
> > `` option `d` stands for [DES](https://docs.dasharo.com/ways-you-can-help-us/#become-a-dasharo-entry-subscription-subscriber) ``
> 
> but it shouldn't be direct link but e.g. `[DES](../ways-you-can-help-us.md#become-a-dasharo-pro-package-subscriber)`

Fixed in 783900a91159a3fca3d0522d3c8a9e7e3aa4ae0e

docs/osf-trivia-list/dts.md
```diff
@@ -40,9 +40,9 @@ This section describes how to do it.
   successful, the message `Verification of the Dasharo DES was successful. They
```
m-iwanicki, line 40: Currently there is no such message:

```
Dasharo DPP credentials have been saved
Checking for Dasharo firmware...
Waiting for network connection ...
Network connection have been established!
Checking if board is Dasharo compatible.
Waiting for network connection ...
Network connection have been established!
A Dasharo Firmware binary has been found for your platform!
```

Should be changed to either `A Dasharo Firmware binary has been found for your platform!` (should probably be fixed in dts-script to say that authorization is ok, somewhat related to https://github.com/Dasharo/dasharo-issues/issues/1063) or maybe say that in case of wrong credentials DTS will print error

m-iwanicki (changes requested):

docs/osf-trivia-list/dts.md
EduKav1813: Added both. The problem has been solved by 9cd1e486637db4a6908efb7c509af04e860a9cb2

EduKav1813: I may also take care of the issue you linked in the future, if I will have time

PR #925, opened 2024-10-25
31M-platform-suspend-and-resume.md: +note about PIKVM unable to wake dut

JanPrusinowski:


docs/unified-test-documentation/dasharo-compatibility/31M-platform-suspend-and-resume.md
```diff
@@ -206,7 +206,8 @@ on the keyboard is tested in this case.
     ```
 
 1. Wait 15 seconds.
-1. Press any key on the keyboard to resume the system.
+1. Press any key on the keyboard to resume the system. (PiKVM does not support
+weaking from suspend state)
```
krystian-hebel, line 210:
```suggestion
1. Press any key on the keyboard to resume the system. (Physical keyboard must be used, PiKVM does not support
waking from suspend state)
```

Maybe also add a link to the PiKVM documentation that mentions that it is disabled by default.

krystian-hebel (changes requested):

PR #946, opened 2024-11-25
Docs on creating Capsule Update releases

philipandag:


mkopec (approved):

docs/kb/capsule-updates-configs.md
```diff
@@ -0,0 +1,73 @@
...
+
+The value looks like `0xMMmmpprr`, where:
+
+- MM - the MAJOR version
+- mm - the MINOR version
+- pp - the PATCH version
+- rr - optional Release Candidate number
+
+For release versions the Release Candidate number is set to `FF` (255 decimal)
+to make sure the version number is higher than any release candidate.
```
SergiiDmytruk, line 56: Actually it's set to `0x80` to have some room for bugfix releases.

mkopec: @philipandag could you submit a follow-up PR for that?

philipandag: > Actually it's set to 0x80 to have some room for bugfix releases.

Right... I should have asked for that first, sorry.
Is releasing hotfixes without changing the release version really desired? Wouldn't that create problems with the consistency and reproducibility of the binaries?

SergiiDmytruk: The release number would be different.  I think it's for things like that coreboot `v24.05.00.01` release.  We might never actually make use of this, but making the room just in case to not regret that we didn't.

philipandag: Updated here: https://github.com/Dasharo/docs/pull/950

krystian-hebel: It is still 0xFF in the example below.

philipandag: Thanks
Fixed the examples here: https://github.com/Dasharo/docs/pull/965

PR #953, opened 2024-12-03
Novacustom building manual: Suggest using 2021 container if build fails

philipandag:
https://github.com/Dasharo/dasharo-issues/issues/1143
Temporary fix. NS5x_TGL v1.5.2 release does not build on 2023 version of the container. Possibly there might be more versions which can't be build according to the instructions.


mkopec (approved):

BeataZdunczyk:
addressed by https://github.com/Dasharo/docs/pull/953

PR #954, opened 2024-12-04
variants/dell_optiplex/recovery.md: add note on flashing stock FW

filipleple:


miczyg1 (changes requested):
I don't see a point in merging that. We already have:

```
If a more serious problem occurs, like
[broken ME](faq.md#cpu-was-replace-warm-reset-required-loop),
it may be necessary to use your
[firmware backup](initial-deployment.md#bios-backup)
and restore content of 8MB chip.
```

just above what you have added. Instead I would do something like this:

```
If a more serious problem occurs, like
[broken ME](faq.md#cpu-was-replace-warm-reset-required-loop),
or you would like to restore Dell firmware,
it may be necessary to use your
[firmware backup](initial-deployment.md#bios-backup)
and restore content of 8MB chip.
```

filipleple:
[Done](https://github.com/Dasharo/docs/compare/09ecc0b43e083a7b6acb7b844392bed667fadc5d..eecf9ceff94b489f1882922b59b40e5930d67fe5#)

docs/variants/dell_optiplex/recovery.md
```diff
@@ -143,8 +143,8 @@ flashrom -w xac -p linux_spi:dev=/dev/spidev1.0,spispeed=16000 -c "MX25L3205D/MX
 ## (Optional) Step 7: Flash 8MB (ME) part
 
 If a more serious problem occurs, like
-[broken ME](faq.md#cpu-was-replace-warm-reset-required-loop),
-it may be necessary to use your
+[broken ME](faq.md#cpu-was-replace-warm-reset-required-loop), or if you would
+like to restore Dell firmware, it may be necessary to use your
 [firmware backup](initial-deployment.md#bios-backup)
 and restore content of 8MB chip.
```
miczyg1, line 149: The whole paragraph needs wrapping now.

miczyg1 (changes requested):

docs/variants/dell_optiplex/recovery.md
filipleple: Right, sorry, [fixed](https://github.com/Dasharo/docs/compare/eecf9ceff94b489f1882922b59b40e5930d67fe5..6c89ec7940996e11a0f9f8bf0657ab36b368013c)

miczyg1 (approved):

PR #955, opened 2024-12-05
Improve test documentation navbar

philipandag:
Change the navigation bar titles in the test specification to match the Dasharo Test & Feature matrix. Append test ID to them to easily distinguish them and sort the entries alphabetically.

PR #956, opened 2024-12-05
Test documentation navbar sorting script

philipandag:
The script can be used to sort specific sections of a text file, like the navbars in documentation.

The script usage:
`<script> [file-to-sort] [start-mark] [end-mark]`
file-to-sort is `mkdocs.yml` by default. The start & end marks are `#pre-commit-sort-start` and `#pre-commit-sort-end` by default, which can be seen in `mkdocs.yml` file.

If we would like to use it then it should definitely be added as a pre-commit hook, not as a loose script. I had problems with creating a working pre-commit hook , so I'll push this script as is for now. 

Merge after https://github.com/Dasharo/docs/pull/955, if at all

*EDIT
Turns out the script can be easily run if it's present in the repo locally. I've changed the .pre-commit-config to run it only on the "mkdocs.yml" file and it works fine. It would be better to move it to a different repository, but I haven't figured out how to do that.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,6 +52,15 @@ repos:
         language: pygrep
         exclude: '^.*.(webp|jpeg|jpg|png|pdf)$'
 
+  - repo: local
+    hooks:
+      - id: sort-mkdocs
+        name: Sort mkdocs nav entries
+        entry: ./sort-mkdocs.py
+        language: system
+        files: "mkdocs.yml"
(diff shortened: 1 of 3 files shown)


macpijan:
Hook integrated here: ab9b3f97cfc4f53548fbd95446605b547b978ca3

and here: https://github.com/3mdeb/hooks/commit/992d64752d382d20c29e1aec9acae1cf6cc53553

PR #1013, opened 2025-03-12
Removing discrepancies between test IDs and other corrections

JakubRed:
Removing discrepancies between test IDs found during vp2430 rc4 testing


miczyg1 (approved):
I don't see any occurrences of DSP002.001 and  DSP002.002

PR #1027, opened 2025-03-26
mkdocs.yml: drop analytics

macpijan:


pietrushnic:
I asked before about fixing this VMware broken link author of VMWare related documentation, which seem to be for Protectli. Can we do something about it to make SEO rank not dropping and linkChecker happy?

pietrushnic (approved):

PR #1035, opened 2025-04-02
fix LinkCheck step in Seo CI

EduKav1813:


EduKav1813:
Now seo pipeline fails if there are any errors in lynchee-parser. Is this a desired behaviour? If yes, I will cleanup the PR to prepare for merge. I added `fail.md` file to test the failing scenario in CI

EduKav1813:
@pietrushnic So the file I added `continue-on-error: false` is not needed, since it would just block all workflows if there is an empty link. This PR if just to test if lychee works correctly on 404 link.

PR #1047, opened 2025-04-18
docs/unified/msi/hcl.md: Move tables to separate files

EduKav1813:
This is POC of moving the hardware compatibility tables out of the main document and then importing them. It is needed so that the https://github.com/Dasharo/dts-hw-conf-gen tool could auto-update those tables.

I am not sure how to properly structure newly created documents to avoid orphan pages though.


requirements.txt
```diff
@@ -45,3 +45,4 @@ virtualenv==20.26.3
 watchdog==4.0.2
 markdown-grid-tables==0.4.0
 mkdocs-macros-plugin==1.3.7
+mkdocs-include-markdown-plugin
```
pietrushnic, line 48: let's hook to specific version

EduKav1813: 7be4c377e06d4008c735c0595357e837347b2400

EduKav1813:
This is POC of  moving the hardware compatibility tables out of the main document and then importing them. It is needed so that the https://github.com/Dasharo/dts-hw-conf-gen tool could auto-update those tables.

I am not sure how to properly structure newly created documents to avoid orphan pages though.

pietrushnic:
Maybe there is no way, and those should be excluded from the link checker? Or somehow marked?

EduKav1813:
@pietrushnic Please see https://github.com/3mdeb/seo-spy/pull/1

EduKav1813:
@pietrushnic Changed the workflow so that the `resources/hcl/` files are ignored by the orphan pages checker.

pietrushnic (approved):
I rendered the site from this branch and things look as expected.

PR #1053, opened 2025-04-29
unified-test-documentation/dasharo-compatibility/31H-usb-type-c.md fi…

JakubRed:
…x: description numeration


BeataZdunczyk (changes requested):
It doesn't render well 
image

PR #1070, opened 2025-06-13
docs/guides/lab-assembly/odroid-assembly-guide.md: Add assembly with …

miczyg1:
…netcard


miczyg1:
@JakubSobota ping

miczyg1:
> @miczyg1 I verified the documentation while assembling the platform and everything seems to be fine, but I wonder why we are not using the appropriate power adapter for the disk, because the one currently installed causes constant problems with the cable reading the platform's power status getting disconnected.

We don't have such adapter, thus the workarounds I made to make the power state still readable.

If you can find such adapter cable, then I may change the documentation.

PR #1105, opened 2025-08-05
docs/variants/dell_optiplex/overview.md: distinguish hardware versions

haeckle:
As discussed in https://github.com/Dasharo/dasharo-issues/issues/1427
Builds and looks fine, no new warning or info labels introduced

I didn't change the picture yet because of
1. copyright - I dont' have the machines so I can't take pictures
2. confusion - the picture in the [doc](https://docs.dasharo.com/variants/dell_optiplex/overview/) and at the [shop](https://shop.3mdeb.com/shop/dasharo-supported-hardware/dasharo-dell-optiplex-7010/) both show the DT version and @pietrushnic [says](https://github.com/Dasharo/dasharo-issues/issues/1182#issuecomment-3073053710) that the
> terminology [here](https://docs.dasharo.com/variants/overview/) is different than Dell, so what we mean by desktop or workstation does not have to reflect what the vendor meant.

but [codewise](https://github.com/Dasharo/coreboot/blob/36244307c6ff2e6e27a43d44702a92011b7ace69/src/mainboard/dell/snb_ivb_workstations/variants/optiplex_9010_sff/overridetree.cb) it's really the SFF version. Are these pictures just show the wrong version all the time?

### DIFF
--- a/docs/variants/dell_optiplex/overview.md
+++ b/docs/variants/dell_optiplex/overview.md
@@ -15,9 +15,33 @@ documentation._
 
 ## Dell OptiPlex 7010/9010 SFF
 
-Dell OptiPlex 7010/9010 SFF is a small SOHO desktop computer, sometimes used
-as a
-[firewall or NAS](https://www.reddit.com/r/homelabsales/comments/uzspg3/comment/iadcyb6/?utm_source=share&utm_medium=web2x&context=3).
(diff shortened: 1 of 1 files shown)


miczyg1:
> Are these pictures just show the wrong version all the time?

I guess so. But I doubt anybody will change it. The efforts on Dell OptiPlex are close to zero right now.

PR #1118, opened 2025-08-21
docs/variants/dell_optiplex: move and update firmware preparation

haeckle:
I tested this on Fedora 41 (with toolbox). As I still don't own the hardware in question I couldn't test if the resulting firmware is working as expected, but the layout looks as following (coreboot with SeaBIOS and TXT support):

Bildschirmfoto vom 2025-08-21 11-45-55

fixes https://github.com/Dasharo/dasharo-issues/issues/1283

### DIFF
--- a/docs/variants/dell_optiplex/building-manual.md
+++ b/docs/variants/dell_optiplex/building-manual.md
@@ -20,8 +20,8 @@ If your choice is to enable TXT support, be advised that there are proprietary
 [ACM blobs](https://doc.coreboot.org/security/intel/acm.html) required for the
 firmware to work properly. They are non-redistributable for the platform in
 question, which means you will need to obtain/extract them yourself and patch
-the result binary using `cbfstool`. The methods are covered later on in the
-[initial deployment guide](initial-deployment.md#firmware-preparation)
+the result binary using `cbfstool`. The methods are covered below in the
+[Include proprietary components](#include-proprietary-components) section.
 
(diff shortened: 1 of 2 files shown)


docs/variants/dell_optiplex/building-manual.md
```diff
@@ -99,8 +99,99 @@ the result binary using `cbfstool`. The methods are covered later on in the
...
+    ```bash
+    make -C util/cbfstool
+    ```
+
+### Obtaining vendor firmware and patching Dasharo
+
+1. For EC and BIOS ACM file the Dell firmware is needed
+
+    ```bash
+    wget --user-agent='Mozilla/4.0 (compatible; MSIE 6.0; Windows NT 5.1; SV1)' https://dl.dell.com/FOLDER05066036M/1/O7010A29.exe
```
miczyg1, line 146: This could use some line wrapping. The URL may be in the next line for example.

miczyg1 (changes requested):
One small comment, otherwise looks good.

docs/variants/dell_optiplex/building-manual.md
haeckle: Right, will do. Should this be a new commit or appended and force pushed? I also got some other small changes (mostly URL fixups)

miczyg1: You have only one commit in the PR, so an amend and force push will be more appropriate I think

miczyg1 (approved):

PR #1141, opened 2025-09-17
Document Dasharo TrustRoot in general and for NovaCustom

SergiiDmytruk:
*ref: ncm-2006*


mkopec:
@SergiiDmytruk LGTM, but precommit is complaining

PR #1144, opened 2025-09-18
docs/osf-trivia-list/dasharo.md: Explain non-LTS release

miczyg1:


docs/osf-trivia-list/dasharo.md
```diff
@@ -147,6 +147,25 @@ There is also a libre, POWER9-based server/workstation Talos II by Raptor
 Computing Systems, which also do not use any binary blobs, however it is more
 expensive than x86 platforms.
 
+## What is Dasharo non-LTS release?
+
+Dasharo non-LTS release is a release that is validated with a reduced set of
+tests. The non-LTS name is adopted from distributions which mark their Long
+Term Support (LTS) releases as more validated/higher quality assurance. For
```
pietrushnic, line 154:
```suggestion
Term Support (LTS) releases as extensively validated/higher quality assurance. For
```

pietrushnic (changes requested):

docs/osf-trivia-list/dasharo.md
```diff
@@ -147,6 +147,25 @@ There is also a libre, POWER9-based server/workstation Talos II by Raptor
 Computing Systems, which also do not use any binary blobs, however it is more
 expensive than x86 platforms.
 
+## What is Dasharo non-LTS release?
```
macpijan, line 150: I am not a huge fan of `non-LTS` name, since everything which is not LTS is essentially non-LTS, leaving us only two options.

pietrushnic: @macpijan I'm also not a fan of it at all. I mentioned that during yesterday's presentation. But creating this kind of prototype brings critical people's attention to the problem and moves the state of the art forward for something I propose for 3+ months. So it is at least a successful strategy.

Some ideas:
* Dasharo Rapid Release - minimal regression only, scope depends on target hardware and configuration, so that it can be 3 or 10 tests, it is very likely <=10% of LTS
* Dasharo Assured release - something in between Rapid and LTS, similar or equal to the "scoped" concept of adding to minimal only tests which validate areas of firmware that experience change,s e.g., USB -> only USB changes, measured boot changes -> only TPM and measured boot
* Dasharo LTS - full scope, typically once a year unless budget is available. We need better communication of EOL.

Propsal: https://github.com/Dasharo/docs/pull/1154

PR #1156, opened 2025-09-23
docs: osf-trivia-list: dts: Add archive with logs to tip

m-iwanicki:
@macpijan should I add whole procedure how to copy those logs (find your storage device, mount it, copy, e.t.c...)?


3mkusiak (approved):
LGTM

PR #1160, opened 2025-09-30
Sv boot wizard rc4

miczyg1:


miczyg1:
linkchecker fails due to shop being down... I checked all the links manually.

PR #1237, opened 2026-03-13
docs(dts): add generated supported-hardware matrix scaffold

WuXieSec:
## Summary
Add an experimental generated hardware mapping matrix for DTS docs to reduce manual drift.

## What this PR adds
- generator script: `scripts/generate_dts_supported_hardware_matrix.py`
- generated include file: `docs/dasharo-tools-suite/documentation/_generated-supported-hardware-matrix.md`
- integration in `supported-hardware.md` via include directive

## Source data
Current generation uses the mapping exported in `Dasharo/dts-configs#39`.

## Why
This is an incremental step for #1576 toward metadata-driven documentation updates, while keeping existing curated content intact.


WuXieSec:
Quick follow-up here. If you want this split into a smaller first step instead of the full scaffold, I can rework it that way.

PR #1240, opened 2026-03-27
docs(ci): add phase-1 build-docs command checks

WuXieSec:
## Summary
Implements a **phase-1** automated test for build documentation commands.

## What this PR adds
- CI workflow: `.github/workflows/docs-build-command-check.yml`
- Checker script: `tools/check_build_docs_commands.py`

## Behavior
- Scans `docs/variants/**/building-manual.md`
- Extracts shell/bash code blocks
- Runs **non-executing** syntax validation (`bash -n`) to catch broken snippets early
- Skips placeholder/pseudocode snippets (`<version>`, `<variant>`, `(docker)`, etc.)
- Emits warnings for uncommon command prefixes (informational, non-blocking)

## Why this approach
For #1153 this provides a low-risk first step that is CI-safe and immediately useful, while keeping room for future phases that execute selected commands in controlled environments.

Related issue: https://github.com/Dasharo/dasharo-issues/issues/1153

### DIFF
--- a/.github/workflows/docs-build-command-check.yml
+++ b/.github/workflows/docs-build-command-check.yml
@@ -1,3 +1,4 @@
+---
 name: docs-build-command-check
 
 on:
--- a/.github/workflows/docs-build-command-check.yml
+++ b/.github/workflows/docs-build-command-check.yml
@@ -1,0 +1,19 @@
+name: docs-build-command-check
+
+on:
+  pull_request:
+  push:
+    branches: [master]
+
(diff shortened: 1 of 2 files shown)


imperatormk:
I'd like to work on this: I have a working implementation that parses build docs, executes in Docker, and compares using romscope. Already tested on QEMU Q35 and by the way it found a CONFIG_LOCALVERSION mismatch on v0.2.1

